PR #31, opened 2024-01-29
add FORCE_FINALITY_AFTER_BLOCKS at the reader/poller level

sduchesneau:
* Added `FORCE_FINALITY_AFTER_BLOCKS` environment variable to override block finality information at the reader/poller level. This allows an operator to pretend that finality is still progressing, N blocks behind HEAD, in the case where a beacon chain fails to do so and is intended as a workaround for deprecated chains like Goerli.


utils/utils.go
```diff
@@ -0,0 +1,25 @@
...
+		if fin64, err := strconv.ParseInt(fin, 10, 64); err == nil {
+			finu64 := uint64(fin64)
+			return &finu64
+		}
+	}
+	return nil
+}
+
+func TweakBlockFinality(blk *pbbstream.Block, maxDistanceToBlock uint64) {
+	if (blk.Number >= maxDistanceToBlock) &&
```
maoueh: Shouldn't you actually have something like 
```suggestion
    blockCountSinceGenesis := blk.Number - bstream.FirstStreamableBlock`
	if (blockCountSinceGenesis >= maxDistanceToBlock) &&
```

utils/utils.go
```diff
@@ -0,0 +1,25 @@
+package utils
```
maoueh: I don't like the `utils` package but unclear what a better name could be. We should at least une `internal/utils` so that it remains outside of the public API of `firehose-core`.

utils/utils.go
sduchesneau: No, it's not that function's role to adjust for the FirstStreamableBlock.
We just don't want to underflow the uint64. 

Clarifying the intent here:
```
func TweakBlockFinality(blk *pbbstream.Block, maxDistanceToBlock uint64) {
	if blk.LibNum > blk.Number {
		panic("libnum cannot be greater than block number")
	}
	if blk.Number < maxDistanceToBlock {
		return // don't touch blocks at the beginning of the chain
	}
	if (blk.Number - blk.LibNum) >= maxDistanceToBlock {
		blk.LibNum = blk.Number - maxDistanceToBlock // force finality
	}
}```

maoueh: Why don't clarify the code also, or maybe you are doing that right now :)

PR #34, opened 2024-02-07
Solana poller integration

billettc:
* Poller is now fetch block in a optimized way, it will fetch several blocks at once and then process them.
* Poller is now handling skipped blocks, it will fetch the next blocks until it find a none skipped block.
* Compare tool is now using dynamic protobuf unmarshaler, it will be able to compare any block type.
* Print tool is now using dynamic protobuf unmarshaler, it will be able to print any block type.
* Print tool is encoding bytes in base64 by default, it can be changed to hex or base58 by using new parameter `bytes-encoding`.


CHANGELOG.md
```diff
@@ -7,6 +7,17 @@ for instructions to keep up to date.
 Operators, you should copy/paste content of this content straight to your project. It is written and meant to be copied over to your project.
 
 If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you should copy the content between those 2 version to your own repository, replacing placeholder value `fire{chain}` with your chain's own binary.
+## UNRELEASED
+
+* Poller is now fetch block in a optimized way, it will fetch several blocks at once and then process them.
```
maoueh, line 12:
```suggestion
* Poller is now fetching block in an optimized way, it will fetch several blocks at once and then process them.
```

CHANGELOG.md
```diff
@@ -40,7 +51,24 @@ If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you s
 
 * Fix `-c ""` not working properly when no arguments are present when invoking `start` command.
 
-* Fix `tools compare-blocks` that would fail on new format.
+* Tools printing Firehose `Block` model to JSON now have `--proto-paths` take higher precedence over well-known types and even the chain itself, the order is `--proto-paths` > `chain` > `well-known` (so `well-known` is lookup last).
```
maoueh, line 54: Those and below are coming from version v1.1.0, remove this.

blockpoller/fetcher.go
```diff
@@ -7,5 +7,6 @@ import (
 )
 
 type BlockFetcher interface {
-	Fetch(ctx context.Context, blkNum uint64) (*pbbstream.Block, error)
+	IsBlockAvailable(requestedSlot uint64) bool
```
maoueh, line 10: Rename to avoid slot since it's Solana centric

blockpoller/init_test.go
```diff
@@ -47,13 +47,17 @@ func (b *TestBlockFetcher) PollingInterval() time.Duration {
 	return 0
 }
 
-func (b *TestBlockFetcher) Fetch(_ context.Context, blkNum uint64) (*pbbstream.Block, error) {
+func (b *TestBlockFetcher) IsBlockAvailable(requestedSlot uint64) bool {
```
maoueh, line 50: Same here

blockpoller/poller.go
```diff
@@ -98,62 +119,178 @@ func (p *BlockPoller) run(resolvedStartBlock bstream.BlockRef) (err error) {
 	}
 }
 
-func (p *BlockPoller) processBlock(currentState *cursor, blkNum uint64) (uint64, error) {
-	if blkNum < p.forkDB.LIBNum() {
-		panic(fmt.Errorf("unexpected error block %d is below the current LIB num %d. There should be no re-org above the current LIB num", blkNum, p.forkDB.LIBNum()))
+func (p *BlockPoller) processBlock(currentState *cursor, block *pbbstream.Block) (uint64, *string, error) {
+	p.logger.Info("processing block", zap.Stringer("block", block.AsRef()), zap.Uint64("lib_num", block.LibNum))
+	if block.Number < p.forkDB.LIBNum() {
+		panic(fmt.Errorf("unexpected error block %d is below the current LIB num %d. There should be no re-org above the current LIB num", block.Number, p.forkDB.LIBNum()))
```
maoueh, line 125: Do we really need to panic here? What if we return simply an `error` here? Or terminating right away the block poller is it implements shutter?

blockpoller/poller.go
```diff
@@ -98,62 +119,178 @@ func (p *BlockPoller) run(resolvedStartBlock bstream.BlockRef) (err error) {
...
+type BlockItem struct {
+	blockNumber uint64
+	block       *pbbstream.Block
+	skipped     bool
+}
+
+func (p *BlockPoller) loadNextBlocks(requestedBlock uint64, numberOfBlockToFetch int) error {
+	p.optimisticallyPolledBlocks = map[uint64]*BlockItem{}
+
+	nailer := dhammer.NewNailer(10, func(ctx context.Context, blockToFetch uint64) (*BlockItem, error) {
```
maoueh, line 181: A comment should be made here to explain that we use the nailer for its linearity of emitting the block in order received.

blockpoller/poller.go
```diff
@@ -35,6 +36,8 @@ type BlockPoller struct {
 	forkDB       *forkable.ForkDB
 
 	logger *zap.Logger
+
+	optimisticallyPolledBlocks map[uint64]*BlockItem
```
maoueh, line 40: We should add a comment here to explain why it's safe/desired behavior to use the block number as a key here. Myself by quickly reading the code I cannot for sure says why.

blockpoller/state_file_test.go
```diff
@@ -33,7 +33,7 @@ func TestFireBlockFinalizer_state(t *testing.T) {
 	fk.AddLink(bstream.NewBlockRef("106a", 106), "105a", &block{Block: blk("106a", "105a", 101)})
 	fk.AddLink(bstream.NewBlockRef("105a", 105), "104a", &block{Block: blk("105a", "104a", 101)})
 	expectedBlocks, reachedLib := fk.CompleteSegment(blk("105a", "104a", 101).AsRef())
-	// simulate firing the blocks
+	// simulate firing the optimisticlyPolledBlocks
```
maoueh, line 36: You search/replace was too broad, please review all renames and adjust them when it's not needed.

chain.go
```diff
@@ -154,13 +154,6 @@ type Chain[B Block] struct {
 }
 
 type ToolsConfig[B Block] struct {
-
-	// SanitizeBlockForCompare is a function that takes a chain agnostic [block] and transforms it in-place, removing fields
-	// that should not be compared.
-	//
-	// The [SanitizeBlockForCompare] is optional, if nil, no-op sanitizer be used.
-	SanitizeBlockForCompare SanitizeBlockForCompareFunc[B]
-
```
maoueh, line 163: Don't remove, those are used in `firehose-ethereum` and also battlefield. Please be really mindful of backward compatibility. We are in a version 1.x so we must keep backward compatibility. Same for `GetSanitizeBlockForCompare`

They should be re-wired too so that the functionality works as expected.

I need those to work.

cmd/tools/compare/tools_compare_blocks.go
```diff
@@ -317,83 +335,72 @@ func (s *state) print() {
...
+		r, err := jd.ReadJsonString(referenceAsJSON)
 		cli.NoError(err, "read JSON reference")
 
-		c, err := jd.ReadJsonString(string(cur))
+		c, err := jd.ReadJsonString(currentAsJSON)
 		cli.NoError(err, "read JSON current")
 
+		//Get the r and the c and print them within a file for further analysis
+
+		referenceWriter, err := os.Create("/Users/arnaudberger/t/reference.json")
```
maoueh, line 382: Please review.

maoueh (changes requested):

PR #35, opened 2024-02-09
Solana block. polling integration

billettc:
* Poller is now fetching blocks in an optimized way, it will fetch several blocks at once and then process them.

* Poller is now handling skipped blocks, it will fetch the next blocks until it find a none skipped block.

* Compare tool is now using dynamic protobuf unmarshaler, it will be able to compare any block type.

* Print tool is now using dynamic protobuf unmarshaler, it will be able to print any block type.

* Print tool is encoding bytes in base64 by default, it can be changed to hex or base58 by using parameter `bytes-encoding`.


blockpoller/poller.go
```diff
@@ -59,37 +62,55 @@ func New(
 	return b
 }
 
-func (p *BlockPoller) Run(ctx context.Context, startBlockNum uint64) error {
+func (p *BlockPoller) Run(ctx context.Context, startBlockNum uint64, numberOfBlockToFetch int) error {
```
sduchesneau: name unclear, rename to blockFetchBatchSize like in firehose-solana

blockpoller/poller.go
```diff
@@ -98,62 +119,178 @@ func (p *BlockPoller) run(resolvedStartBlock bstream.BlockRef) (err error) {
...
+func (p *BlockPoller) fetchBlockWithHash(blkNum uint64, hash string) (*pbbstream.Block, error) {
+	p.logger.Info("fetching block with hash", zap.Uint64("block_num", blkNum), zap.String("hash", hash))
+	_ = hash //todo: hash will be used to fetch block from  cache
+
+	p.optimisticallyPolledBlocks = map[uint64]*BlockItem{}
+
+	var out *pbbstream.Block
+	var skipped bool
+	err := derr.Retry(p.fetchBlockRetryCount, func(ctx context.Context) error {
+		//todo: get block from cache
```
sduchesneau, line 286: what's missing here to do that ?

blockpoller/poller_test.go
```diff
@@ -157,7 +157,7 @@ func TestForkHandler_run(t *testing.T) {
 			f := New(blockFetcher, blockFinalizer)
 			f.forkDB = forkable.NewForkDB()
 
-			err := f.run(tt.startBlock)
+			err := f.run(tt.startBlock, 1)
```
sduchesneau, line 160: shouldn't we write a test for the optimistic poller too ? seems tricky enough to deserve a test

cmd/tools/compare/tools_compare_blocks.go
```diff
@@ -317,83 +334,54 @@ func (s *state) print() {
...
 
-		r, err := jd.ReadJsonString(string(ref))
+		r, err := jd.ReadJsonString(referenceAsJSON)
 		cli.NoError(err, "read JSON reference")
 
-		c, err := jd.ReadJsonString(string(cur))
+		c, err := jd.ReadJsonString(currentAsJSON)
 		cli.NoError(err, "read JSON current")
 
 		if diff := r.Diff(c).Render(); diff != "" {
```
sduchesneau, line 379: who wrote that ? seems strangely named

sduchesneau, line 379: see last comment

CHANGELOG.md
```diff
@@ -28,10 +40,16 @@ If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you s
 
 * The `tools print one-block` now works correctly on blocks generated by omni-chain `firecore` binary.
 
+* Tools printing Firehose `Block` model to JSON now have `--proto-paths` take higher precedence over well-known types and even the chain itself, the order is `--proto-paths` > `chain` > `well-known` (so `well-known` is lookup last).
+
+* The `tools print one-block` now works correctly on blocks generated by omni-chain `firecore` binary.
+
 * The various health endpoint now sets `Content-Type: application/json` header prior sending back their response to the client.
 
 * The `firehose`, `substreams-tier1` and `substream-tier2` health endpoint now respects the `common-system-shutdown-signal-delay` configuration value meaning that the health endpoint will return `false` now if `SIGINT` has been received but we are still in the shutdown unready period defined by the config value. If you use some sort of load balancer, you should make sure they are configured to use the health endpoint and you should `common-system-shutdown-signal-delay` to something like `15s`.
```
sduchesneau: duplicates in here!

chain.go
```diff
@@ -154,13 +154,6 @@ type Chain[B Block] struct {
 }
 
 type ToolsConfig[B Block] struct {
```
sduchesneau, line 156: why did we lose the sanitization option ? I feel that it may have been useful in some cases

sduchesneau (changes requested):
good work, just a few minor changes please!

chain.go
maoueh: I need it for Ethereum compare because `failureReason` are different so we exclude them as part of the compare.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/streamingfast/firehose-core
 
-go 1.21
+go 1.22
```
maoueh, line 3: We won't be able to release using `sfreleaser` for now as the `goreleaser-cross` Docker image with Go 1.22 is not published yet (although 3 days ago a PR was merged to bring it in, so it seems to be coming).

sduchesneau: hmm ok I'll try to release it under 1.21 for now then !!!

chain.go
sduchesneau: ok I added it back with a new format that will work. minimal adjustments in fireeth will be needed.

PR #37, opened 2024-02-26
add api key authentication to NewFirehoseFetchClient

fschoell:


firehose/client/client.go
```diff
@@ -77,10 +77,10 @@ func (a *ApiKeyAuth) RequireTransportSecurity() bool {
 	return true
 }
 
-func NewFirehoseFetchClient(endpoint, jwt string, useInsecureTSLConnection, usePlainTextConnection bool) (cli pbfirehose.FetchClient, closeFunc func() error, err error) {
+func NewFirehoseFetchClient(endpoint, jwt, apiKey string, useInsecureTSLConnection, usePlainTextConnection bool) (cli pbfirehose.FetchClient, closeFunc func() error, callOpts []grpc.CallOption, err error) {
```
maoueh, line 80: This is a breaking change on `firehose-core`, so let's use find another way to bring that in for now. 

Also, before we were using `DialOption` but I see you switched to using call options and ensuring they are properly passed later. Why can't we keep the old signature?

maoueh (changes requested):

firehose/client/client.go
fschoell: basically I just adopted the FetchClient to the changes to the StreamClient Stephane did in here: https://github.com/streamingfast/firehose-core/commit/7411f2b49205b5b9fe785302c509f8ca877955f9

Seems like he just forgot the FetchClient when adding the api key authentication? Or maybe there are other reasons to not have it the same way on the FetchClient, but having consistency is probably good here.

maoueh: Ok I see, let me check with him as this is definitely a breaking change.

maoueh:
Will merge as-is as it probably don't affect users that much. Will try to take care more care of the public APIs in the future.

PR #42, opened 2024-03-21
add --merger-delete-threads

fschoell:
the S3 API of Ceph has performance issues with file deletes, so it is necessary to increase the number of parallel delete threads. This adds a `--merger-delete-threads` flag for the merger that customises that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you s
 
 ## Unreleased
 
-* Added `--merger-delete-threads` to customize the number of threads the merger will use to delete files.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Operators, you should copy/paste content of this content straight to your projec
 
 If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you should copy the content between those 2 version to your own repository, replacing placeholder value `fire{chain}` with your chain's own binary.
 
+## Unreleased
+
+* Added `--merger-delete-threads` to customize the number of threads the merger will use to delete files.
```
maoueh: Can you add a note that this is particularly useful for those on Ceph storage provider.

maoueh: And how manu threads you use there

maoueh (changes requested):
The rest is all good

CHANGELOG.md
fschoell: added some context

maoueh (approved):

PR #63, opened 2024-09-04
add reader-node-firehose

fschoell:
This adds a new reader-node app that reads blocks from another Firehose endpoint. This allows indexers to bootstrap a Firehose/Substreams stack without having to set up an instrumented blockchain node initially.

### DIFF
--- a/node-manager/app/firehose_reader/app.go
+++ b/node-manager/app/firehose_reader/app.go
@@ -1,0 +1,146 @@
+// Copyright 2019 dfuse Platform Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package firehose_reader
+
+import (
+	"fmt"
+	"github.com/streamingfast/bstream/blockstream"
+	pbbstream "github.com/streamingfast/bstream/pb/sf/bstream/v1"
+	dgrpcserver "github.com/streamingfast/dgrpc/server"
+	dgrpcfactory "github.com/streamingfast/dgrpc/server/factory"
+	nodeManager "github.com/streamingfast/firehose-core/node-manager"
+	"github.com/streamingfast/firehose-core/node-manager/mindreader"
+	"github.com/streamingfast/logging"
+	pbheadinfo "github.com/streamingfast/pbgo/sf/headinfo/v1"
+	"github.com/streamingfast/shutter"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+)
+
+type Config struct {
+	GRPCAddr                   string
+	OneBlocksStoreURL          string
+	OneBlockSuffix             string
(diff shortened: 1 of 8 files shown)


node-manager/app/firehose_reader/app.go
```diff
@@ -0,0 +1,151 @@
...
+		a.zlogger,
+		a.tracer,
+	)
+	if err != nil {
+		return err
+	}
+
+	a.zlogger.Debug("configuring shutter")
+	mindreaderLogPlugin.OnTerminated(a.Shutdown)
+	a.OnTerminating(mindreaderLogPlugin.Shutdown)
```
maoueh, line 117: Ideally, we would avoid completely the mindreader plugin, there is overall little reason to use outside avoiding re-implementing two things:
- Save received one block to one-blocks-store
- Emit block on block stream server

I understand it make the implementation easier. But I have the overall feeling that it will make a much simpler implementation since we will move away from the node-manager.

We should also move to something else than `node-manager` as the top-level package.

node-manager/app/firehose_reader/console_reader.go
```diff
@@ -0,0 +1,144 @@
...
+
+	stream, err := f.firehoseClient.Blocks(context.Background(), &pbfirehose.Request{
+		StartBlockNum:   int64(startBlock),
+		Cursor:          string(cursor),
+		StopBlockNum:    stopBlock,
+		FinalBlocksOnly: false,
+	}, f.callOpts...)
+	if err != nil {
+		return fmt.Errorf("failed to request block stream from Firehose: %w", err)
+	}
```
maoueh, line 77: How do we deal with connection errors?

node-manager/app/firehose_reader/reader_stats.go
```diff
@@ -0,0 +1,55 @@
...
+		lastBlock: pbbstream.BlockRef{},
+		blockRate: dmetrics.MustNewAvgRateFromPromCounter(BlockReadCount, 1*time.Second, 30*time.Second, "blocks"),
+	}
+}
+
+func (s *firehoseReaderStats) StartPeriodicLogToZap(ctx context.Context, logger *zap.Logger, logEach time.Duration) {
+	ctx, s.cancelPeriodicLogger = context.WithCancel(ctx)
+
+	go func() {
+		ticker := time.NewTicker(logEach)
```
maoueh, line 29: ticker has a `Stop()` let's defer stop it.

maoueh: `blockRate` too has a `Stop` method. Could go in `StopPeriodicLogToZap()` too.

maoueh (changes requested):

node-manager/app/firehose_reader/app.go
fschoell: Sounds good. Should I move it into its own top-level package or is there another package to include it in?

sduchesneau:
1. @fschoell  let's add this new app to `appsNotRunningByDefault` in cmd/apps/start.go

2. I agree with @maoueh that using the consoleReader feels very hackish and I'd much prefer copying the bits that you need to reimplement something more slim.

With these fixed, I think we could merge it.

node-manager/app/firehose_reader/app.go
maoueh: A new top-level `reader-node-firehose` seems good, the `package <name>` could be shorter if needed.

matthewdarwin:
Fred will continue on this in the new year.

maoueh:
@matthewdarwin I reviewed the SQL PRs too. If there is something I missed, let me know.

maoueh:
Merged on the side.

PR #64, opened 2024-09-06
fix metering endpoint for Firehose block streams

fschoell:


firehose/server/server.go
```diff
@@ -79,7 +79,7 @@ func New(
 
 		meter := dmetering.GetBytesMeter(ctx)
 		auth := dauth.FromContext(ctx)
-		metering.Send(ctx, meter, auth.UserID(), auth.APIKeyID(), auth.RealIP(), auth.Meta(), "sf.firehose.v2.Firehose/Block", response)
+		metering.Send(ctx, meter, auth.UserID(), auth.APIKeyID(), auth.RealIP(), auth.Meta(), "sf.firehose.v2.Firehose/Blocks", response)
```
maoueh, line 82: This will change metering dashboards we have. Wondering if this would affects other providers too.

We could coordinate that for our own dashboards, not sure for the community. Minimally, the CHANGELOG will need to be enhanced to describe the extent that this change could cause.

maoueh (changes requested):

firehose/server/server.go
fschoell: Ah seems like the PR isn't clear, `sf.firehose.v2.Firehose/Blocks` was the endpoint send in metering events, until it was recently changed in here (and impacted our dashboards): https://github.com/streamingfast/firehose-core/commit/5782c51a7054aa716d76cd08d9e9f724237fec92#diff-279f3c1dcecbee568a2ae6fbeebbb130a7f2e7bceb7cf1b0f69a2d2b9fb00cf3R82

I assume this was done on accident so this PR reverts it back to the original behaviour.

PR #67, opened 2024-11-12
Refactor blockpoller initialization to use first streamable block

billettc:
Replaced start block with first streamable block in BlockPoller to ensure accurate initiation. Added a check for the state file's existence and integrated state initialization with fetching the first streamable block. Updated tests accordingly.


blockpoller/state_file.go
```diff
@@ -33,20 +33,36 @@ type stateFile struct {
...
 }
 
+func (p *BlockPoller) isStateFileExist(stateStorePath string) bool {
+	if stateStorePath == "" {
+		p.logger.Info("No state store path set, skipping cursor check")
+		return false
+	}
+	fp := filepath.Join(stateStorePath, "cursor.json")
+	_, err := os.Stat(fp)
+	exist := err == nil || os.IsExist(err)
```
maoueh: I think `IsExists(err)` could be true only on doing a write, but `Stat` does a read only. It should be only `err == nil`

PR #69, opened 2024-12-09
Refactored how `firecore tools` handles printing

maoueh:
There is now a common flag for all `tools` subcommand `output` that can be one of: text, json, jsonl, protojson or protojsonl.

Commands that had flags defined have been removed, this should be backward compatible.

The text printer is now more clever and is able to print details about most blocks.

### DIFF
--- a/cmd/tools/check/blocks.go
+++ b/cmd/tools/check/blocks.go
@@ -2,7 +2,6 @@ package check
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -181,6 +180,8 @@ func validateBlockSegment[B firecore.Block](
 		return
 	}
 
+	printer := print2.TextOutputPrinter{}
+
 	seenBlockCount := 0
 	for {
 		block, err := readerFactory.Read()
@@ -231,7 +232,7 @@ func validateBlockSegment[B firecore.Block](
 			seenBlockCount++
 
 			if printDetails == PrintStats {
-				err := print2.PrintBStreamBlock(block, false, os.Stdout)
+				err := printer.PrintTo(block, os.Stdout)
 				if err != nil {
 					fmt.Printf("❌ Unable to print block %s: %s\n", block.AsRef(), err)
 					continue
@@ -239,6 +240,12 @@ func validateBlockSegment[B firecore.Block](
 			}
 
 			if printDetails == PrintFull {
+				printer, err := print2.GetOutputPrinter(globalToolsCheckCmd, chain.BlockFileDescriptor())
+				if err != nil {
+					fmt.Printf("❌ Unable to create output printer: %s\n", err)
+					break
+				}
+
 				var b = chain.BlockFactory()
 
 				if _, ok := b.(*pbbstream.Block); ok {
@@ -251,14 +258,11 @@ func validateBlockSegment[B firecore.Block](
 					break
 				}
 
-				out, err := json.MarshalIndent(b, "", "  ")
-
+				err = printer.PrintTo(b, os.Stdout)
 				if err != nil {
 					fmt.Printf("❌ Unable to print full block %s: %s\n", block.AsRef(), err)
 					continue
 				}
-
-				fmt.Println(string(out))
 			}
 
 			continue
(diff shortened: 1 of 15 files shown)


cmd/tools/check/blocks.go
```diff
@@ -231,14 +232,20 @@ func validateBlockSegment[B firecore.Block](
...
-				err := print2.PrintBStreamBlock(block, false, os.Stdout)
+				err := printer.PrintTo(block, os.Stdout)
 				if err != nil {
 					fmt.Printf("❌ Unable to print block %s: %s\n", block.AsRef(), err)
 					continue
 				}
 			}
 
 			if printDetails == PrintFull {
+				printer, err := print2.GetOutputPrinter(globalToolsCheckCmd, chain.BlockFileDescriptor())
```
maoueh, line 243: I plan to remove all printing support from `check`.

cmd/tools/tools.go
```diff
@@ -29,13 +30,37 @@ import (
...
+			JSON and JSONL have the caveat to print enum value using the integer value
+			instead of the name which would be more convenient.
+
+			ProtoJSON and ProtoJSONL being able to print only Protobuf messages, they
+			are refused on commands that are not returning Protobuf messages.
+
+			One of: text, json, jsonl, protojson, protojsonl
+		`))
+
+		flags.String("bytes-encoding", "hex", "Encoding for bytes fields when printing in 'text', 'json' or 'jsonl' --output, either 'hex', 'base58' or 'base64'")
```
sduchesneau, line 61: idea for this:
maybe LONG byte arrays should always be printed in hex, and this `bytes-encoding` flag should always be for byte arrays below 64 bytes or something...

One of the issue is that base58 is excruciatingly slow when dealing with long byte arrays

cmd/tools/print/printer.go
```diff
@@ -0,0 +1,131 @@
...
+		if printer == "protojson" {
+			indent = "  "
+		}
+
+		return NewProtoJSONOutputPrinter(indent, registry), nil
+	}
+
+	if printer == "text" {
+		// Supports the `transactions` flag defined on `firecore tools print` sub-command,
+		// we should move it to a proper `text` sub-option like `output-text-details` or something
```
sduchesneau, line 70: maybe just a different output mode:
`short`
`full`

or `summary` vs `fulltext`

json/marshallers.go
```diff
@@ -197,3 +198,31 @@ func ToBase64(encoder *jsontext.Encoder, t []byte, options json.Options) error {
...
+}
+
+func EncodeHex(bytes []byte) string {
+	return hex.EncodeToString(bytes)
+}
+
+// Encode encodes the given bytes using the specified encoding.
+func Encode(bytesEncoding string, bytes []byte) string {
+	switch {
+	case strings.EqualFold(bytesEncoding, "base58"):
```
sduchesneau, line 247: here we could switch based on the length, as mentionned earlier.. just an idea but I think it would make sense

sduchesneau (approved):
Good job !
2 suggested small changes, but not required to go forward.

cmd/tools/check/blocks.go
sduchesneau: agreed.

cmd/tools/compare/tools_compare_blocks.go
```diff
@@ -73,9 +73,7 @@ func NewToolsCompareBlocksCmd[B firecore.Block](chain *firecore.Chain[B]) *cobra
 
 	flags := cmd.PersistentFlags()
 	flags.Bool("diff", false, "When activated, difference is displayed for each block with a difference")
-	flags.String("bytes-encoding", "hex", "Encoding for bytes fields, either 'hex' or 'base58'")
```
sduchesneau, line 76: Here you replaced with the global `bytes-encoding` that supports more values (like base64), but the implementation in tools_compare_blocks.go does NOT support anything else than `hex` or `base58`, watch out there needs to be a few changes later in this file, in `func Compare(...)`

cmd/tools/tools.go
maoueh: Seems rather hard to understand if encoding switches based on length IMO. And since it's for printing, not sure it's a big thing with the performance, we are not in Solana ;)

cmd/tools/print/printer.go
maoueh: Yeah let's discuss that part, I created a generic block printer using Golang protoreflect, we could maybe extend it!

json/marshallers.go
maoueh: The byte encoding should have been an enum from the start, would make everything better. Let's refactor it later.

PR #92, opened 2025-04-23
fix: prevent mutex lock leakage in bundler when MergeAndStore fails

joncode5:
This commit fixes a deadlock issue in the merger bundler when processing blocks that are ahead of the current bundle boundary. The previous implementation would acquire a mutex lock before calling MergeAndStore(), but if this operation failed, the function would return early without releasing the lock, preventing other operations from proceeding.

The fix encapsulates the critical section in a closure with a deferred unlock, ensuring the mutex is always released regardless of error conditions. This prevents lock leakage and potential deadlocks during error scenarios.


joncode5:
@maoueh please take a look

PR #96, opened 2025-05-11
Added dynamic fetching of networks from The Graph's network registry

GabrielCartier:
Removed old types and switched everything to Pinax network type.


firehose/info/info_filler.go
```diff
@@ -13,37 +13,37 @@ var DefaultInfoResponseFiller = func(firstStreamableBlock *pbbstream.Block, resp
...
 		}
 	}
 
 	if !validate {
 		if resp.ChainName == "" {
 			// still try to fill the chain name if it is not given
-			if chain := wellknown.WellKnownProtocols.ChainByGenesisBlock(firstStreamableBlock.Number, firstStreamableBlock.Id); chain != nil {
-				resp.ChainName = chain.Name
+			if chain := wellknown.ChainByGenesisBlock(firstStreamableBlock.Number, firstStreamableBlock.Id); chain != nil {
+				resp.ChainName = chain.ShortName
```
GabrielCartier: @maoueh this is not exactly like it was in the code before (like `Ethereum` vs. `ethereum`). 
Does this have an impact? Because we check shortname, fullname and aliases in ChainByName, I'm guessing it's ok?

well-known/chains.go
```diff
@@ -1,352 +1,95 @@
...
+		}
+		if slices.Contains(network.Aliases, name) {
+			return network
 		}
 	}
 	return nil
 }
+
+// BlockIdEncodingForNetwork returns the InfoResponse_BlockIdEncoding for a given network based on its Firehose.BytesEncoding.
+func BlockIdEncodingForNetwork(network *registry.Network) pbfirehose.InfoResponse_BlockIdEncoding {
```
GabrielCartier, line 77: Should this be here or somewhere else?

cmd/apps/start.go
```diff
@@ -84,26 +85,48 @@ func start[B firecore.Block](cmd *cobra.Command, dataDir string, args []string,
...
+	aliases := sflags.MustGetStringSlice(cmd, "advertise-chain-aliases")
+	blockIdEncoding := sflags.MustGetString(cmd, "advertise-block-id-encoding")
+
+	blockIDEncodingEnum := pbfirehose.InfoResponse_BLOCK_ID_ENCODING_UNSET
+	firstStreamableBlock := bstream.GetProtocolFirstStreamableBlock
+
+	// If --advertise-chain-name is set, but any of the other advertise flags are not, fill them from the registry
+	if cmd.Flags().Changed("advertise-chain-name") &&
+		(!cmd.Flags().Changed("advertise-chain-aliases") ||
+			!cmd.Flags().Changed("advertise-block-id-encoding")) {
```
GabrielCartier, line 97: @maoueh do we also want to check `advertise-block-features` here too? I don't think this is something we can extract from the registry but I might be wrong.

cmd/apps/start.go
```diff
@@ -84,26 +85,48 @@ func start[B firecore.Block](cmd *cobra.Command, dataDir string, args []string,
...
+
+		network := networks.GetRegistryNetworksWithFirehose().Find(chainName)
+		if network != nil {
+			if !cmd.Flags().Changed("advertise-chain-aliases") {
+				aliases = network.Aliases
+			}
+			if !cmd.Flags().Changed("advertise-block-id-encoding") {
+				blockIDEncodingEnum = info.BlockIdEncodingForNetwork(network)
+			}
+			firstStreamableBlock = uint64(network.Genesis.Height)
```
GabrielCartier: I'm not 100% sure how that was working previously because `bstream.GetProtocolFirstStreamableBlock` is always returning 0, but seems like in the registry some networks are not 0. Either way, there's no flag for this but thought it made sense to add it if we got the network.

firehose/info/info_filler.go
```diff
@@ -4,46 +4,49 @@ import (
...
 	"strings"
 
+	registry "github.com/pinax-network/graph-networks-libs/packages/golang/lib"
 	pbbstream "github.com/streamingfast/bstream/pb/sf/bstream/v1"
-	wellknown "github.com/streamingfast/firehose-core/well-known"
 	pbfirehose "github.com/streamingfast/pbgo/sf/firehose/v2"
+	"github.com/streamingfast/substreams/networks"
 )
 
 var DefaultInfoResponseFiller = func(firstStreamableBlock *pbbstream.Block, resp *pbfirehose.InfoResponse, validate bool) error {
```
GabrielCartier, line 13: I'm not sure implementation is correct here. Because we do some prefetching/prepopulating in `start.go`,  it feels like some of the code is redundant.

Like, why do re-set the `resp.BlockIdEncoding` if it's already set? I guess same with `resp.ChainName` and `resp.ChainNameAliases`

Let me know about it @maoueh.

cmd/apps/start.go
sduchesneau: No, the features are either provided, or discovered from the blocks themselves

cmd/apps/start.go
sduchesneau: Hmm this will be an issue: the first-streamable-block is ALWAYS defined from the user flag `--common-first-streamable-block`. It is the one that sets the value in `bstream.GetProtocolFirstStreamableBlock`, that is used throughout the code.

The INFO endpoint must serve that one, and not the one that comes from the network.

ex: mainnet starts at block 0,
MY endpoint starts at block 22_000_000 (set by the flag ...)
we want the info endpoint to serve 22M.

cmd/apps/start.go
```diff
@@ -84,26 +85,48 @@ func start[B firecore.Block](cmd *cobra.Command, dataDir string, args []string,
...
 	infoServer := info.NewInfoServer(
-		sflags.MustGetString(cmd, "advertise-chain-name"),
-		sflags.MustGetStringSlice(cmd, "advertise-chain-aliases"),
-		blockIDEncoding,
+		chainName,
+		aliases,
+		blockIdEncoding,
 		sflags.MustGetStringSlice(cmd, "advertise-block-features"),
-		bstream.GetProtocolFirstStreamableBlock,
+		firstStreamableBlock,
```
sduchesneau: keep the bstream.GetProtocolFirstStreamableBlock.

Ignore the one from the registry.

firehose/info/info_filler.go
sduchesneau: You're right, a few things:

1) the blockIDEncoding and ChainNameAliases should NOT be replaced if they are already set
2) the ChainName DOES need to be set with the value from the registry (there's a comment in the code regarding canonical name...)
3) Do not use `chain.ShortName` for the name, instead, use `chain.ID` everywhere.
we want `solana-mainnet-beta` for chain solana, instead of `Solana`.

cmd/apps/start.go
sduchesneau: Test like this:

`firecore tools download-from-firehose mainnet.sol.streamingfast.io:443 0:100 sol-blocks`
`go run ./cmd/firecore start firehose -c "" --common-merged-blocks-store-url=$(pwd)/sol-blocks --common-live-blocks-addr= --common-first-streamable-block=100 --advertise-chain-name=solana`

then `grpcurl --plaintext   localhost:10015    sf.firehose.v2.EndpointInfo/Info` 
should give you the same as this:
```
{
  "chainName": "solana-mainnet-beta",
  "chainNameAliases": [
    "solana",
    "solana-mainnet"
  ],
  "firstStreamableBlockNum": "100",
  "firstStreamableBlockId": "DdwYftNVqFsFUAfSqD59nT356KeSDfVJkUHjsGdvjR8y",
  "blockIdEncoding": "BLOCK_ID_ENCODING_BASE58"
}
```
check the firstStreamableBlockNum here
(maybe more aliases exist from the registry, no worries bout that)

sduchesneau (changes requested):
not bad, a few changes needed.
I included the commands on how to test it for solana, for example.

cmd/apps/start.go
GabrielCartier: That makes sense, fixed it.

GabrielCartier:
rebased on develop and adjusted from the review.

sduchesneau (approved):
lgtm

PR #98, opened 2025-06-06
fix well-known types generator, update firehose-beacon types

fschoell:
This PR updates the firehose-beacon proto to include the new Electra spec. It also comes with some minor fixes

* removes the https:// url schema from the buf request (the request fails if it's included)
* deduplicates well known protos (by name and data)
* ensures the networks are sorted for easier review when regenerating the types


proto/generator/generator.go
```diff
@@ -46,12 +49,20 @@ func main() {
 	)
 
 	var protofiles []ProtoFile
+	registeredNetwork := make([]*registry.Network, 0, len(networks.GetFirehoseRegistry()))
```
sduchesneau: pluralise slices!

PR #110, opened 2026-02-17
feat(blockpoller): poll blocks in parallel

TPXP:
This PR improves the block poller by enabling it to poll multiple blocks in parallel, thus greatly improving throughput.

It also enhances logging to report RPC call failures, which were previously silently failing


billettc (approved):

billettc (approved):

maoueh:
It seems most of this was already done via 1a2066d differently. 

I'm working on a follow up PR for polling for multiple providers when doing backfill, this will be open as another PR.

PR #149, opened 2026-06-06
index-builder: include block number, ID and payload type in unmarshalling errors

YaroShkvorets:
## Problem

When the `index-builder` app hits a block whose payload fails to unmarshal, it dies with:

```
Fatal error in app index-builder:
proto: cannot parse invalid wire-format data
```

There is no indication of **which block** or which merged-blocks bundle is at fault, forcing the operator to `fireeth tools check merged-blocks -f` the whole store to locate it. We hit this in production on a MegaETH deployment (single corrupted block deep inside an otherwise-healthy bundle).

## Change

Wrap the payload unmarshalling error in the index-builder block handler with the block number, block ID and payload type URL:

```
unmarshalling block #6258830 (d82cd86e…) payload (type.googleapis.com/sf.ethereum.type.v2.Block): proto: cannot parse invalid wire-format data
```

The block number directly identifies the offending bundle file.


maoueh:
If rebased, checks should pass now for external contributions

sduchesneau (approved):
LGTM

PR #152, opened 2026-06-09
rpc: don't hold clients lock across the fetch call

YaroShkvorets:
Required by https://github.com/streamingfast/firehose-ethereum/pull/159

## Problem

`WithClientsContext` held `clients.lock` for the **entire duration of the fetch callback `f`**, not just while selecting a client:

```go
clients.lock.Lock()
defer clients.lock.Unlock()
...
v, err := f(ctx, client)   // seconds of network I/O, lock still held
```

The mutex is only meant to guard the rolling-strategy state (which endpoint to hand out next, for failover). Holding it across `f` means **every concurrent caller serializes** on it.

The block poller's parallel-prefetch path (`blockpoller` → `dhammer.Nailer`, used when `blockFetchBatchSize > 1`) runs each fetch through `WithClients`, so all prefetch workers queued on this single lock. Result: "parallel" block fetching ran strictly one-at-a-time, and a `block-fetch-batch-size` knob had no effect. Especially painful on chains with large blocks 

## Fix

Scope the lock to **only** the rolling-strategy selection, releasing it before calling `f`:

```go
clients.lock.Lock()
clients.rollingStrategy.reset()
client, err := clients.rollingStrategy.next(clients)
clients.lock.Unlock()
...
v, err := f(ctx, client)   // runs concurrently across callers
```

Also makes `BlockPoller.fetching` an `atomic.Bool` (`CompareAndSwap`): it was read in the run loop and written in the spawned prefetch goroutine without synchronization — a pre-existing data race flagged by `go test -race` once the concurrent path is exercised. CAS is race-clean and guarantees exactly one prefetch goroutine in flight.

## Ordering is preserved

Blocks are fired strictly in ascending order regardless of fetch completion order: `dhammer.Nailer` linearizes output in input order, and the poll loop fires one block at a time from the single run goroutine.

## Measured impact

End-to-end through the real `reader-node` pipeline (megaETH, one-block files/min):

| `block-fetch-batch-size` | throughput |
|---|---|
| 1 (current) | ~0.5 blk/s |
| 10 | ~3.3 blk/s (**6.7x**) |
| 16 | ~4.6 blk/s |
| 32 | ~5.7 blk/s |
| 64 | ~7.1 blk/s |

At batch=1 the poller can't keep up with the ~1 blk/s chain; batch>=8 keeps up and catches up. Near-linear to ~16, then diminishing returns.

## Tests

- `rpc.TestWithClients_doesNotSerialize` — 4 concurrent callers must all be inside `f` at once; fails (times out) if the lock is ever re-held across `f`.
- `blockpoller.TestBlockPoller_parallelFetchPreservesOrder` — `blockFetchBatchSize=10` with out-of-order fetch completion; asserts blocks still fire in strict ascending order.

Both pass under `go test -race`.

### DIFF
--- a/blockpoller/poller_parallel_test.go
+++ b/blockpoller/poller_parallel_test.go
@@ -1,0 +1,64 @@
+package blockpoller
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"testing"
+	"time"
+
+	pbbstream "github.com/streamingfast/bstream/pb/sf/bstream/v1"
+	"github.com/streamingfast/firehose-core/rpc"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+)
+
+// linChainFetcher is a thread-safe fetcher for a simple linear chain. It is used
+// to exercise parallel block fetching (blockFetchBatchSize > 1). It deliberately
+// delays LOWER block numbers MORE than higher ones, so the concurrent fetches
+// complete out of order — the poller must still fire blocks to the handler in
+// strict ascending order.
+type linChainFetcher struct {
+	libNum   uint64
+	maxBlock uint64
+
+	mu      sync.Mutex
+	fetched int
+}
+
+func (f *linChainFetcher) IsBlockAvailable(uint64) bool { return true }
+
+func (f *linChainFetcher) Fetch(ctx context.Context, c any, num uint64) (*pbbstream.Block, bool, error) {
+	// higher block number => shorter delay => completes earlier (scrambled completion order)
+	time.Sleep(time.Duration(f.maxBlock-num+1) * 5 * time.Millisecond)
+	f.mu.Lock()
+	f.fetched++
+	f.mu.Unlock()
+	return blk(fmt.Sprintf("%da", num), fmt.Sprintf("%da", num-1), f.libNum), false, nil
+}
+
+func TestBlockPoller_parallelFetchPreservesOrder(t *testing.T) {
+	const first, lib, stop = uint64(100), uint64(100), uint64(110)
+
+	expect := make([]*pbbstream.Block, 0, stop-first)
+	for n := first; n < stop; n++ {
+		expect = append(expect, blk(fmt.Sprintf("%da", n), fmt.Sprintf("%da", n-1), lib))
+	}
+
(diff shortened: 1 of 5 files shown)


rpc/clients_concurrent_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"go.uber.org/zap"
+)
+
+// TestWithClients_doesNotSerialize guards the fix in WithClientsContext: the
+// clients lock must only protect rolling-strategy selection, NOT the call to
+// f. If the lock is (re)introduced around f, concurrent callers serialize and
+// only one can be inside f at a time — this test then deadlocks until the
+// per-entry timeout fires and fails.
+func TestWithClients_doesNotSerialize(t *testing.T) {
```
maoueh, line 24: Go has synctest package now to test async code, let's use it.

blockpoller/poller_parallel_test.go
```diff
@@ -0,0 +1,64 @@
...
+func (f *linChainFetcher) Fetch(ctx context.Context, c any, num uint64) (*pbbstream.Block, bool, error) {
+	// higher block number => shorter delay => completes earlier (scrambled completion order)
+	time.Sleep(time.Duration(f.maxBlock-num+1) * 5 * time.Millisecond)
+	f.mu.Lock()
+	f.fetched++
+	f.mu.Unlock()
+	return blk(fmt.Sprintf("%da", num), fmt.Sprintf("%da", num-1), f.libNum), false, nil
+}
+
+func TestBlockPoller_parallelFetchPreservesOrder(t *testing.T) {
```
maoueh, line 42: This should also probably

maoueh (changes requested):

maoueh (approved):

maoueh (approved):

maoueh:
Please rebase, I've fixed the CI problem.

YaroShkvorets:
rebased

YaroShkvorets:
Looks like Docker Scout fails without credentials

image

maoueh:
@YaroShkvorets Yeah sorry, if you could rebase again, PRs checks should be fixed now.

YaroShkvorets:
Rebased. Still fails on that "Docker Scout" step

matthewdarwin:
yay!

PR #158, opened 2026-06-11
fix(firehose): serve first streamable block from hub in Fetch/Block

maoueh:
## Summary

`sf.firehose.v2.Fetch/Block` hangs forever — no response, no error — when a client requests the chain's **first streamable block** on a freshly started instance. The independent `Stream/Blocks` long-running stream is unaffected. After enough blocks have been produced the same request succeeds instantly, which is what made this look intermittent.

One-line fix: in the single-block handler, compare the requested block number against the hub's lowest retained block with `>=` instead of `>`.

## Root cause

`firehose/block_getter.go` routes a single-block request like this:

```go
// check for block in live segment: Hub
if g.hub != nil && num > g.hub.LowestBlockNum() {
    if blk := g.hub.GetBlock(num, id); blk != nil { return blk, nil }
    return nil, status.Error(codes.NotFound, "live block not found in hub")
}
// else: fall through to merged-blocks store ...
```

`hub.LowestBlockNum()` resolves (via `Forkable.LowestBlockNum` → `ForkDB.CompleteSegment`) to the **lowest block still retained** in the forkable segment — that block *is* present in the hub and *is* serveable via `GetBlock`.

The guard uses a strict `>`, so a request for exactly the lowest retained block is **excluded from the hub branch** and falls through to the merged-blocks store. On a freshly started chain:

- The lowest retained block is the **first streamable block** (e.g. block 1 with `first-streamable-block=1`).
- No merged bundle (`0000000000`) has been flushed yet — the merger needs a full 100-block bundle, and the chain has only produced a handful of blocks.
- `bstream.FetchBlockFromMergedBlocksStore` drives a `FileSource` that, on a missing merged file, **retries every second indefinitely** (`filesource.go` `launchReader`, the `!exists` branch — `errorOutOnMissingFile` is not set for point fetches). It never returns, so the RPC never responds.

Because only the single lowest block trips the off-by-one (e.g. `num=2` at head 30 gives `2 > 1` → hub → instant), the symptom is specific to the *first* streamable block.

### Why it "fixes itself" later

At head ~544 the forkable has purged blocks below `LIB - keptBlocksBelowLIB`, so block 1 is no longer in the hub and `1 > 42` correctly skips it — but by then the merged bundle `0000000000` has been flushed, so the merged-store path finds block 1 immediately. Both regimes are self-consistent; the hang only exists in the early window where the block is in the hub but the guard refuses to look.

## Live confirmation (pprof)

One stuck goroutine per hung Fetch, on the streamingfast logging/pprof server (`127.0.0.1:1065`):

```
curl -s 'http://localhost:1065/debug/pprof/goroutine?debug=2' | grep -A20 FetchBlockFromMergedBlocksStore
```

```
bstream.(*FileSource).run                 filesource.go:298   [select, blocked]
bstream.FetchBlockFromMergedBlocksStore   single_block_fetcher.go:109  (num=1)
firehose.(*BlockGetter).Get.func1         block_getter.go:74
firehose.(*BlockGetter).Get               block_getter.go:73
firehose/server.(*Server).Block           blocks.go:51        ← Fetch entry
```

## The fix

`firehose/block_getter.go`: `num > g.hub.LowestBlockNum()` → `num >= g.hub.LowestBlockNum()`. The lowest retained block is now served from the hub, which holds it. This also fixes the same hang for genesis block 0 on `first-streamable-block=0` chains.

## Note / possible follow-up (bstream)

This PR fixes the trigger and restores correct behavior. A defensive backstop is still worth doing separately in `bstream`: `FetchBlockFromMergedBlocksStore` is a point fetcher, yet on a missing merged file its `FileSource` retries every second **indefinitely**, ignoring context cancellation. That unbounded wait should be bounded.

The intent is *not* to fail fast — waiting for a bundle that is about to be flushed is a legitimate use. Rather, cap the wait with a **long timeout (e.g. 1–2h)**; once it elapses, return an error so the **caller decides** whether it still wants to keep waiting for the bundle to arrive, instead of a single fetch goroutine blocking forever. Not required to fix this issue.

## Testing

- `go test ./firehose/...` — pass
- `go build ./firehose/...` — pass
- **E2E regression test** added (`relayer/block_getter_e2e_test.go`, `TestFetchFirstStreamableBlockServedFromHub`): boots a `dummy-blockchain` reader-node (testcontainers), feeds a live `ForkableHub`, then drives a `firehose.BlockGetter` at the hub's lowest retained block with an empty merged-blocks store and nil forked store — the fresh-chain state where the block lives only in the hub. A wall-clock guard fails the test if `Get` hangs. Verified it **passes** with the `>=` fix and **fails** (`Get hung on the first streamable block`) when reverted to strict `>`. Skip-gated behind `TEST_E2E` / `CI`, matching the existing relayer e2e test.


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/firehose-core:24da760-amd64</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/firehose-core:24da760-amd64</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:db9d5dbbbcf1ed9fdbe58fdce80ab047c1d2eaab491120266dab949d2c7270eb</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
<tr><td>size</td><td>161 MB</td></tr>
<tr><td>packages</td><td>505</td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>
<li><code>c136481b8f4cd58cb213a22ee358ea1047bc963a338c61cdccb518e233409f86</code></li>

<li><code>noble</code></li>

<li><code>noble-20260509.1</code></li>
</ul>
</td></tr>
<tr><td>digest</td><td><code>sha256:023f8a753c22258c9fe2d0005a7d28258038da7d620e9f93e9ad78aa266f9f11</code></td></tr>
<tr><td>vulnerabilities</td><td>critical: 0 high: 1 medium: 26 low: 11 </td></tr>
</table>
</details>

<table></table>

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you s
 
 ### Fixed
 
+- Firehose: fix `sf.firehose.v2.Fetch/Block` hanging forever when requesting the first streamable block on a freshly started chain. The single-block handler used a strict `>` comparison against the hub's lowest retained block, so a request for exactly that block (the first streamable block at startup) skipped the live hub and fell through to the merged-blocks store, where it waited indefinitely for a merged bundle that had not been flushed yet. The comparison is now `>=`, so the lowest retained block is served from the hub.
```
maoueh:
```suggestion
- Firehose: fix `sf.firehose.v2.Fetch/Block` hanging until merged bundle is created when requesting the first streamable block on a freshly started chain. The single-block handler used a strict `>` comparison against the hub's lowest retained block, so a request for exactly that block (the first streamable block at startup) skipped the live hub and fell through to the merged-blocks store, where it waited indefinitely for a merged bundle that had not been flushed yet. The comparison is now `>=`, so the lowest retained block is served from the hub.
```

sduchesneau (approved):
LGTM

PR #172, opened 2026-07-02
merger: configurable one-block-files prune distance

YaroShkvorets:
The one-block-files pruner hardcoded the prune distance to the bundle size (`100`), so merged one-block files are deleted ~100 blocks behind the merge frontier. A relayer/firehose whose hub falls behind by more than that can't find the one-block files to bridge the gap (the hub has no merged-store fallback), and gets stuck reconnecting — eventually dying with `cannot link block after reconnection, restart required`.

Happened on Hypercore (~14 blocks/s): the relayer fell behind, the ~100-block window is only seconds wide at that rate, and the needed one-block files were already merged-and-deleted. Restarts sometimes helped, sometimes didn't. Raising the prune distance keeps them around long enough to self-recover.

## Change

Add `--merger-prune-one-block-files-after` (default `100`), mirroring `--merger-prune-forked-blocks-after`. Used by `startOldFilesPruner`; forked-blocks pruner untouched. Clamped to a minimum of the bundle size (`100`).

Safe by construction: raising it only makes the pruner more conservative (deletes fewer/older files, never not-yet-merged ones). Default behavior unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you s
 
 ## Unreleased
 
+### Added
+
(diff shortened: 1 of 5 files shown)


YaroShkvorets:
Could replace with env var if adding cli flag is too much

sduchesneau (approved):
LGTM. These chains block rates are crazy 😓

sduchesneau:
@YaroShkvorets will make it to next release, not expecting release until next week tho.

PR #174, opened 2026-07-06
Fix merger lock leak, dead circuit breaker, skip-loop boundary, pipe leak and pruner shutdown

sduchesneau:
## Fixes

- **Dead circuit breaker** (`merger/merger.go`): the error branch reset
  `consecutiveErrors` before incrementing it, so the `>= 10` shutdown threshold was
  unreachable and the merger retried forever on persistent store failure while
  healthz stayed SERVING. The reset now only happens on success.
- **Bundler lock leak** (`merger/bundler.go`): returning `errTerminating` from the
  skip-forward loop left the bundler mutex held, deadlocking pruners in
  `getSafeBaseBlockNum()`. The lock is now released via `defer`, and `eg.Stop()` is
  checked before marking a bundle in-flight so termination no longer leaves a stale
  in-flight entry.
- **Skip-loop off-by-one** (`merger/bundler.go`): a bundle covers
  `[base, base+size)`, so a post-gap block landing exactly on `base+size` belongs to
  the next bundle; the `>` condition attributed it to the previous one, producing a
  merged file whose only block is out of range plus a duplicate in the next bundle.
  Condition changed to `>=`.
- **Pipe reader leak** (`merger/merger_io.go`, `merger/bundlereader.go`): when
  `WriteObject` failed or timed out without draining the streaming bundle reader,
  the feeding goroutine blocked forever in `io.Copy` holding an open one-block
  reader, leaking per retry. The pipe reader is now closed on every attempt, and the
  reader runs under the per-attempt timeout ctx so its between-files check fires.
- **Pruners ignore shutdown** (`merger/merger.go`): both pruner goroutines looped on
  `time.Sleep` and kept deleting files during/after shutdown. They now select on
  `Terminating()` vs the timer.

## Tests

New regression tests, each verified to fail against the pre-fix code:

- `TestMergerRun_ConsecutiveWalkErrorsTriggerShutdown`
- `TestBundlerProcessBlockTerminatingReleasesLock`
- `TestBundlerSkipLoopBoundary` (table: exact boundary, one past, exact next-bundle start)
- `TestMergerIO_WriteObjectErrorClosesStreamReader`
- `TestPrunersStopOnShutdown`

`go test ./merger/...` passes. `go test -race ./merger/...` passes except the
pre-existing data race inside `TestMergerRun_FailedBundleCausesShutdown`'s own test
code (unsynchronized `err`/counter access), which fails identically on clean
`develop`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you s
 ### Fixed
 
 - `tools print merged-blocks <store>` no longer truncates its output when a merged-blocks file is missing (open range with no explicit stop, or a bounded range extending past the last available file): with the bumped `bstream` dependency it now prints every available block first, instead of discarding in-flight files on an async shutdown. On an open range it caps the stream at the last available merged-blocks file (found with an `O(log n)` existence probe rather than a full store listing) so it stops cleanly instead of erroring on the expected-missing next file. Also fixes an off-by-one that stopped one block early on a closed range, and a potential unsigned underflow when a closed range ends at block 0.
+- Merger: fixed a lock leak, a dead consecutive-errors circuit breaker, a skip-loop boundary off-by-one, a streaming bundle-reader goroutine leak, and pruners that ignored shutdown.
 
 ## v1.15.0
 
--- a/merger/bundler.go
+++ b/merger/bundler.go
@@ -259,7 +259,9 @@ func (b *Bundler) ProcessBlock(_ *pbbstream.Block, obj interface{}) error {
 	lastBlock := b.irreversibleBlocks[len(b.irreversibleBlocks)-1]
 	b.irreversibleBlocks = []*bstream.OneBlockFile{lastBlock, obf}
 	b.baseBlockNum += b.bundleSize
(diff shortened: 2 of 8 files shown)


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/firehose-core:bb4a425-amd64</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/firehose-core:bb4a425-amd64</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:312323b6fb7940d2cd02d7d9fd806fbc87cc1cd06212a767f5b2279d0367ca2d</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
<tr><td>size</td><td>163 MB</td></tr>
<tr><td>packages</td><td>507</td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>
<li><code>noble</code></li>

</ul>
</td></tr>
<tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 14 low: 5 </td></tr>
</table>
</details>

<table></table>

sduchesneau (commented):
LGTM!

billettc (approved):

PR #189, opened 2026-07-17
relayer: seed head_block_number metric from hub head on startup

YaroShkvorets:
## Problem

The relayer's `head_block_number` Prometheus gauge is a labeled gauge (`GaugeVec` keyed by `app`), so the `head_block_number{app="relayer"}` series only appears in `/metrics` after `.Set()` is called at least once. That call only happens when a **live** block flows through the forkable:

- `relayer.go` wires `forkable.WithMetrics(metrics.HeadBlockNumber, ...)`.
- The gauge is written only `if p.liveMetrics` (forkable), and `liveMetrics` is flipped only in `hub.ProcessBlock` (`SetLiveMetrics()`), which runs only when a live block arrives.

On chains that **mine blocks on demand**, no new block may arrive after a relayer restart, so the series is never created and scrapers can't see it — even though the hub already knows its head from bootstrapping the one-block files (`hub.bootstrap()` populates the forkable and closes `Ready`, but does not touch `liveMetrics`).

## Fix

Seed `metrics.HeadBlockNumber` from `r.hub.HeadNum()` immediately after the hub becomes ready, so consumers can read the known head right away instead of waiting for the next mined block. It stays static until a live block arrives, which is correct for an idle on-demand chain.

Note: this only helps when the hub bootstrapped from one-block files (the common restart case). If the one-blocks store is empty, the relayer is genuinely not-ready and has no head to report until its first live block — unchanged by this PR.

### DIFF
--- a/relayer/relayer.go
+++ b/relayer/relayer.go
@@ -144,9 +144,17 @@ func (r *Relayer) Run() {
 	zlog.Info("waiting for hub to be ready...")
 	<-r.hub.Ready
 
-	// Set head_block_number so we know the head immediately
-	if headNum := r.hub.HeadNum(); headNum != 0 {
+	// Seed head metrics from the bootstrapped hub head so we report it
+	// immediately, instead of only once a live block flows through the forkable.
+	if headNum, headID, headTime, _, err := r.hub.HeadInfo(); err == nil {
+		zlog.Info("seeding head metrics from hub head",
+			zap.Uint64("head_num", headNum),
+			zap.String("head_id", headID),
+			zap.Time("head_time", headTime),
+		)
 		metrics.HeadBlockNumber.SetUint64(headNum)
+		metrics.HeadBlockTimeDrift.SetBlockTime(headTime)
+		metrics.HeadBlockRelativeDrift.SetLastBlock(headTime)
 	}
 
 	r.OnTerminating(func(e error) {
--- a/relayer/relayer.go
+++ b/relayer/relayer.go
@@ -144,6 +144,11 @@ func (r *Relayer) Run() {
 	zlog.Info("waiting for hub to be ready...")
 	<-r.hub.Ready
 
+	// Set head_block_number so we know the head immediately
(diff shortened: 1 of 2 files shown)


relayer/relayer.go
```diff
@@ -144,6 +144,16 @@ func (r *Relayer) Run() {
...
 	<-r.hub.Ready
 
+	// The head_block_number gauge is otherwise only set once a live block flows
+	// through the forkable (see hub.ProcessBlock -> SetLiveMetrics). On chains
+	// that produce blocks on demand, no new block may arrive after a restart, so
+	// the metric would never materialize even though the hub already knows its
+	// head from bootstrapping the one-block files. Seed it here so consumers
+	// (e.g. monitoring) can read the known head right away.
+	if headNum := r.hub.HeadNum(); headNum != 0 {
+		metrics.HeadBlockNumber.SetUint64(headNum)
```
maoueh, line 155: I think we should set all metrics here in one show. Metrics code would normally do on each block:

```
		// we have longest chain and we just sent it, we update drift metrics
		if p.metricsHeadBlockNum != nil {
			p.metricsHeadBlockNum.SetUint64(blk.Number)
		}
		if p.metricsHeadTimeDrift != nil {
			p.metricsHeadTimeDrift.SetBlockTime(blk.Time())
		}
		if p.metricsRelativeBlockDrift != nil {
			p.metricsRelativeBlockDrift.SetLastBlock(blk.Time())
		}
```

We can use `HeadInfo()` to gather `headNum uint64, headID string, headTime time.Time` so we can re-do the same thing (which makes me wonder if the fix should be in the hub directly which once boostrapped, could set the metrics itself).

maoueh (changes requested):

maoueh (approved):

relayer/relayer.go
YaroShkvorets: Right. Changed to setting all 3 in the same place.

PR #190, opened 2026-07-22
Add finalized_block_number metric

maoueh:
## Why

Firehose apps expose `head_block_number{app}`, but nothing on the drift dashboard says how far behind finality that head is. This adds `finalized_block_number{app}` so a dashboard can show head↔finality drift in blocks as `head_block_number - finalized_block_number`, per app.

Block number only: `pbbstream.Block.LibNum` is on every block, so the finalized *number* is available wherever a head number is set today. The finalized *time* is not — `LibNum` carries no timestamp and resolving it needs a forkdb lookup by ID, which only relayer and reader-node could ever satisfy, and only while the LIB block is still in the window. Block number is the finality signal we trust today; seconds-behind-finality can be approximated on the dashboard as blocks × chain block time.

## What

**Metric definition** — new `metrics/` package (`coremetrics`), following the `relayer/metrics/source_drift.go` precedent rather than adding to `dmetrics`: a package-level `GaugeVec` labelled by `app`, registered once at init time, with per-app instances holding only the label value. That registration shape is what keeps it safe when several apps run in one process (`fire{chain} start reader-node,relayer,merger`); registering the collector per `dmetrics.Set` would panic on a duplicate descriptor.

**Reported by:**

| App | Source |
|---|---|
| reader-node, reader-node-stdin, reader-node-firehose | `block.LibNum`, through the node-manager metrics manager |
| firehose | `blk.LibNum` on the block stream handler |
| relayer | `blk.LibNum` on the forkable live path, seeded from the hub head at startup |
| merger, block-indexer | same value as head (see below) |

**Merger and index-builder report finalized == head.** Both only ever process final blocks (index-builder streams with `FinalBlocksOnly`, merger bundles irreversible one-block files), so their head is finalized by construction. They deliberately do *not* publish the LIB number carried by those blocks: that value trails their own head and would make the drift panel read as if the merger were lagging finality.

**No breaking change for chain repositories.** `NewMetricsAndReadinessManager` takes the new metric through a variadic `WithFinalizedBlockNumberMetric(...)` option rather than a sixth positional argument, and `firehose.Modules` gets an optional, nil-guarded field.

**bstream** — the relayer's head metrics are set inside the forkable, so the finalized number is set on the same path via the new `forkable.WithFinalizedBlockNumMetric` option (streamingfast/bstream#61, merged). Computing it relayer-side instead (a ticker, or the per-source handler) would report head and finalized from different observation points.

## Out of scope

- substreams-tier1 / tier2: the head metric is set inside the substreams library.
- `finalized_block_time_drift`: see the reasoning above.

## Tests

- `metrics/finalized_block_test.go`: metric name, `app` label, per-app values.
- `node-manager/monitor_test.go`: the metrics manager reports `LibNum` when the option is given, and reports nothing when it is not.
- bstream side: value set from `LibNum` on the live path, not set when live metrics are off.

The `relayer` e2e test fails locally on `develop` as well (the `dummy-blockchain` container exits 1), unrelated to this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,16 @@ Operators, you should copy/paste content of this content straight to your projec
 
 If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you should copy the content between those 2 version to your own repository, replacing placeholder value `fire{chain}` with your chain's own binary.
 
+## Unreleased
+
+### Added
+
+- New `finalized_block_number{app}` metric, reported by `reader-node`, `reader-node-stdin`, `reader-node-firehose`, `relayer`, `firehose`, `merger` and `block-indexer`. Paired with the existing `head_block_number{app}`, it lets dashboards show how far behind finality each app's head is, as `head_block_number - finalized_block_number`. The `merger` and `block-indexer` only ever process final blocks, so they report the same value for both.
+
+### Changed
+
+- Bumped `bstream` for `forkable.WithFinalizedBlockNumMetric`, the option backing the new `finalized_block_number` metric on the `relayer`.
+
 ## v1.16.0
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you s
 
 - Bumped `bstream` for `forkable.WithFinalizedBlockNumMetric`, the option backing the new `finalized_block_number` metric on the `relayer`.
 
+### Security
+
+- Bumped `google.golang.org/grpc` to `v1.82.1`, fixing GHSA-hrxh-6v49-42gf (HIGH, CVSS 8.8), an uncaught exception reachable from the network.
+
 ## v1.16.0
 
 ### Added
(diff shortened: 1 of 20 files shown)


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/firehose-core:8fe24a9-amd64</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/firehose-core:8fe24a9-amd64</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:a7a9ebd519d0758221bb9ae6d5d4c3c0f63b64d511fe01bc6dcfff5a712fab9b</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
<tr><td>size</td><td>163 MB</td></tr>
<tr><td>packages</td><td>507</td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>
<li><code>noble</code></li>

</ul>
</td></tr>
<tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 14 low: 5 </td></tr>
</table>
</details>

<table></table>

billettc (approved):

PR #198, opened 2026-08-03
rpc: named providers and explicit failback reset for the Clients pool

maoueh:
## Why

sf-operator is adding a QuickNode fallback RPC provider to the `aval-mainnet` Firehose reader, which runs `fireeth tools poller optimism <rpc-endpoint> <first-streamable-block>`. firehose-ethereum is getting a repeatable `--provider=<url>#<name>` flag in a sibling change; this is the firehose-core side that makes provider names usable at failure time.

Two gaps in the current pool:

1. `Clients[C]` stored only `clients []C`. `eth-go`'s `rpc.Client` keeps its endpoint unexported with no accessor, so once a client is in the pool nothing could say which provider it is — `WithClientsContext` appended every attempt error with no attribution.
2. `StickyRollingStrategy` was sticky forever. `reset()` only zeroed `usedClientCount` and `next()` then returned `prevIndex()`, i.e. the last client used. A single transient error on the primary moved the caller to the fallback permanently until process restart. The only escape was `StartSorting`, whose goroutine reset the strategy through an inline type switch — bundled with sorting, so callers that don't want sorting couldn't get the reset.

## What changed

New public API on `rpc.Clients`:

```go
func (c *Clients[C]) Add(client C)                    // unchanged behavior; auto-names "client-<idx>"
func (c *Clients[C]) AddNamed(client C, name string)  // new
func (c *Clients[C]) Names() []string                 // new, in pool order
func (c *Clients[C]) Reset()                          // new: re-prefer declared order
```

- `Clients` tracks a provider name per client. `Add` keeps working unchanged for the existing callers across the chain repos and auto-names the client `client-<index>`. `Sort` reorders the names alongside the clients so `Names()` stays truthful after a sort.
- `Reset()` brings the rolling strategy back to the pool's declared order — the failback operation. The reset now lives behind the `RollingStrategy` interface as `resetToDeclaredOrder()`, so the type switch in `StartSorting` is gone and `StartSorting` calls `Reset()`.
- `RollingStrategy.next` returns `(C, int, error)` so `Clients` can resolve the selected client's name. The interface's methods are unexported, so it can only be implemented inside the package — not a breaking change for external callers.
- `WithClientsContext` wraps each failed attempt as `fmt.Errorf("provider %q: %w", name, err)` before appending to the multierror, and logs each roll at `warn` level with `from_provider`/`to_provider` so a failover is visible in logs without waiting for the final error. A given `from -> to` roll is logged at `warn` at most once every 30s (review feedback): a pool using `RollingStrategyAlwaysUseFirst` rolls on every call, so a permanently down first provider would otherwise warn once per RPC call. The rolls in between go to `debug` and are counted as `suppressed_rolls` on the next warning.
- `NewClients` tolerates a nil logger.

No metrics counter: the `rpc` package has no metrics registry and wiring one in would mean new dependencies and plumbing, so it was skipped rather than invented.

## Tests

New coverage in `rpc/clients_named_test.go`: named vs auto-named clients, names surviving a sort, error attribution text for both naming modes, the roll log's from/to fields, the roll-log sampling (per `from -> to` pair, with the suppressed count), and the failback case — that `Reset()` sends the next call back to the declared-order primary — for both strategies. Test loggers follow the `logging.PackageLogger` pattern (new `rpc/log_test.go`), and the existing `rpc` tests were moved off `zap.NewNop()`.

`./rpc/...` and `./blockpoller/...` pass, `go build ./...` is clean. `relayer/TestRelayerReconnectsAfterSourceRestart` fails, but it fails on a clean `develop` checkout too and is unrelated to this change.

## Follow-up

The sibling firehose-ethereum change needs a firehose-core release to `go get`; the entries here sit under a new `## Unreleased` CHANGELOG section (latest tag is `v1.16.1`, this is additive so `v1.17.0`).

### DIFF
--- a/rpc/clients.go
+++ b/rpc/clients.go
@@ -14,9 +14,9 @@ import (
 
 var ErrorNoMoreClient = errors.New("no more clients")
 
-// defaultRollLogSamplingInterval is how often at most a given `from -> to` roll is
-// logged at `warn` level, see [Clients.logRoll].
-const defaultRollLogSamplingInterval = 30 * time.Second
+// rollLogInterval is how often at most a given `from -> to` roll is logged at `warn`
+// level, see [Clients.logRoll].
+const rollLogInterval = 30 * time.Second
 
 type Clients[C any] struct {
 	clients               []C
@@ -26,13 +26,8 @@ type Clients[C any] struct {
 	lock                  sync.Mutex
 	logger                *zap.Logger
 
-	rollLogSamplingInterval time.Duration
-	rollLogs                map[string]*rollLogState
-}
-
-type rollLogState struct {
-	lastLoggedAt time.Time
-	suppressed   int
+	// rollLoggedAt is the last time a given `from -> to` roll was logged at `warn` level.
+	rollLoggedAt map[string]time.Time
 }
 
 func NewClients[C any](maxBlockFetchDuration time.Duration, rollingStrategy RollingStrategy[C], logger *zap.Logger) *Clients[C] {
@@ -45,8 +40,7 @@ func NewClients[C any](maxBlockFetchDuration time.Duration, rollingStrategy Roll
 		rollingStrategy:       rollingStrategy,
 		logger:                logger,
 
-		rollLogSamplingInterval: defaultRollLogSamplingInterval,
-		rollLogs:                map[string]*rollLogState{},
+		rollLoggedAt: map[string]time.Time{},
 	}
 }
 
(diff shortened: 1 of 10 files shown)


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/firehose-core:9a7f7b7-amd64</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/firehose-core:9a7f7b7-amd64</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:7efa1159824efcd82d22a6eba4f6d56b5de16c57e4251c7e5a5508699b6f4590</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
<tr><td>size</td><td>174 MB</td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>
<li><code>noble</code></li>

</ul>
</td></tr>
<tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 17 low: 5 </td></tr>
</table>
</details>

<table></table>

rpc/clients.go
```diff
@@ -76,15 +130,27 @@ func WithClientsContext[C any, V any](clients *Clients[C], ctx context.Context,
...
+			}
 			clients.lock.Unlock()
-			if err != nil {
-				errs = multierror.Append(errs, err)
+			if nextErr != nil {
+				errs = multierror.Append(errs, nextErr)
 				return v, errs
 			}
 
+			clients.logger.Warn("rolling to next RPC provider",
```
sduchesneau: this can spam logs with RollingStrategyAlwaysUseFirst if the first provider is down (one WARN every rpc call).

We should sample

rpc/rolling_strategy.go
```diff
@@ -40,22 +57,24 @@ func (s *StickyRollingStrategy[C]) next(clients *Clients[C]) (client C, err erro
...
-		client = clients.clients[s.nextClientIndex]
+		index = s.nextClientIndex
+		client = clients.clients[index]
 		s.usedClientCount = s.usedClientCount + 1
 		s.nextClientIndex = s.nextClientIndex + 1
-		return client, nil
+		return client, index, nil
 	}
 
 	if s.nextClientIndex == len(clients.clients) { //roll to 1st client
```
sduchesneau: this is unreachable: line 54 will zero the nextClientIndex

sduchesneau (approved):
Ok if you merge https://github.com/streamingfast/firehose-core/pull/199 and address my two other comments, as discussed !

rpc/clients.go
maoueh: Sampled: a given `from -> to` roll now logs at `warn` at most once every 30s, the ones in between drop to `debug` and are counted as `suppressed_rolls` on the next warning. Covered by `TestWithClientsRollLogIsSampled` (5 rolls of a down primary -> 1 warning) and `TestWithClientsRollLogPerProviderPair` (sampling is per pair, a roll to a different provider is not swallowed). 9aafc50

rpc/rolling_strategy.go
maoueh: Right, dead code — the check above already zeroes `nextClientIndex` and the just-been-reset branch returns before reaching it. Removed in 9aafc50.

PR #199, opened 2026-08-03
rpc: sort a snapshot of the clients pool under lock

sduchesneau:
Targets `feature/rpc-clients-named-providers` (#198), not `develop`.

## Why

Two problems in `Sort`, one of which #198 makes worse.

**1. Data race.** `Sort` reads `clients.clients` — and, since #198, `clients.names` via `nameAt` — without holding `clients.lock`; the lock is only taken for the final two assignments. `StartSorting` runs `Sort` on its own goroutine, concurrently with `Add`/`AddNamed` and with the client selection in `WithClientsContext`. `append` in `Add` can reallocate the backing array and rewrite the slice header while `Sort` reads it, with no happens-before edge. #198 adds a second racing read on the new `names` slice, and does it through `nameAt`, whose own doc comment says *"the caller is responsible of holding the lock"*.

**2. Lost writes.** `Sort` is a read-modify-write across a gap that spans one network call per client. A client added during that gap is silently overwritten out of existence by the final assignment.

## Why not just take the lock for the whole function

Because `Sort` has no timeout. `firehose-solana/cmd/firesol/rpc/fetcher.go:76` — the only production caller of `StartSorting` — passes `cmd.Context()`, the root context, and `Sort` never applies `maxBlockFetchDuration`. `FetchSortValue` is one network round-trip per client, run sequentially.

Holding the lock across that loop means every `WithClientsContext` blocks for N round-trips on each sort interval, and if one provider hangs, the lock is held until it returns — **every block fetch in the pool wedged for the duration**. A single slow provider would take down the whole pool, the exact failure the fallback-provider work in #198 exists to prevent.

## What changed

Snapshot under the lock, fetch unlocked, then publish under the lock only if the pool length still matches the snapshot — otherwise redo the round.

The length is an exact staleness check here: the pool only ever grows through `Add`/`AddNamed` or gets permuted by `Sort` itself, so an unchanged length means no client was added and `sorted` covers the whole pool. Two concurrent `Sort`s keep the length and hold the same *set*, so the loser overwrites the order but never the membership, and the next round settles it. No generation counter needed.

The retry bails on a cancelled context so it cannot spin during shutdown. In practice it fires at most once: pools are built at startup, before `StartSorting`.

`nameAt` drops out of `Sort` and is once again only called under the lock, as its comment requires. No public API change, no new fields or methods.

## Tests

New `rpc/sort_concurrent_test.go` — all three fail against #198's `rpc/sort.go`, pass with the fix:

- `TestSortConcurrentWithPoolAccess` — four concurrent sorters against a goroutine driving `Names`/`WithClients`; the race detector reports 7 `DATA RACE` findings on the unfixed code.
- `TestSortKeepsClientAddedMidRound` — a fetcher that adds a client mid-round; asserts it is not merely kept but sorted into place by the redone round.
- `TestSortStopsRetryingOnCancelledContext` — retry stops on shutdown.

`go test -race -count=2 ./rpc/... ./blockpoller/...` passes; `go build ./...`, `go vet ./rpc/...`, `gofmt -l rpc/` clean.

## Not addressed

Separate review notes on #198, left out to keep this diff to the locking issue:

- With `RollingStrategyAlwaysUseFirst`, the new roll `warn` fires once per `WithClientsContext` call when the primary is down — per block, in `blockpoller`. The roll-per-call behavior is pre-existing; only the log is new.
- `Clients.Reset()` landing while `WithClientsContext` has released the lock around `f` can re-select the provider that just failed and restarts the retry budget.
- `Add`'s auto-name `client-<len>` no longer matches position after a sort, and can collide with an existing name.
- `rolling_strategy.go` has an unreachable second `if s.nextClientIndex == len(clients.clients)` branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,12 @@ If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you s
 
 - `rpc.WithClients`/`rpc.WithClientsContext` now attribute each failed attempt to its provider, the returned error reads `provider "<name>": <error>`, and each roll to another provider is logged at `warn` level with the `from_provider`/`to_provider` names.
 
+### Fixed
+
+- `rpc.Sort` no longer reads the clients pool without holding the lock, which was a data race against `Add`/`AddNamed` and against the client selection in `WithClientsContext`, since `StartSorting` runs `Sort` on its own goroutine. It now sorts a snapshot taken under the lock, keeping the per-client network fetches lock-free.
+
+- A client added to the pool while `rpc.Sort` was running is no longer dropped. `Sort` snapshots the pool, fetches a sort value per client over the network, then writes the pool back, and that write used to overwrite anything appended in between. It now redoes the round instead of publishing, so the new client is sorted along with the rest.
+
 ## v1.16.1
 
 ### Added
--- a/rpc/rolling_strategy_test.go
+++ b/rpc/rolling_strategy_test.go
@@ -13,6 +13,7 @@ type rollClient struct {
 	callCount int
 	name      string
 	sortValue uint64
+	pool      *Clients[*rollClient] // only set by tests that reach back into the pool
 }
 
 func (r *rollClient) fetchSortValue(_ context.Context) (sortValue uint64, err error) {
--- a/rpc/sort.go
+++ b/rpc/sort.go
@@ -2,6 +2,7 @@ package rpc
 
 import (
 	"context"
+	"slices"
 	"sort"
 )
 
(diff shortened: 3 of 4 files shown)


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/firehose-core:a8c1c88-amd64</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/firehose-core:a8c1c88-amd64</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:c225e1c7f4969cd5bcaa20fb354e8464914e231c35306d7b8240a50ee0360961</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
<tr><td>size</td><td>174 MB</td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>
<li><code>noble</code></li>

</ul>
</td></tr>
<tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 17 low: 5 </td></tr>
</table>
</details>

<table></table>

maoueh (approved):

PR #208, opened 2026-08-11
Bump substreams to latest develop + fix superviser restart deadlock

maoueh:
## Substreams bump

Bumps `github.com/streamingfast/substreams` to `v1.21.1-0.20260811030713-864626c02e63` (latest `develop`).

- [substreams#866](https://github.com/streamingfast/substreams/pull/866) — `substreams-tier1` now logs the parallel-worker negotiation. The `incoming Substreams Blocks request` entry carries a `parallelism` object (requested, granted by the authentication layer, effective count and which of the two capped it, plan tier, stage layer executors) plus `parallel_segment_count` and `stage_count`.
- `firehose-networks` bump to v0.2.3 on the substreams side, matching the one already vendored here.

## Superviser deadlock fix

CI on this branch hung for the full 10 minutes test timeout in `TestSuperviser_CanBeRestartedCorrectly`. The failure is **not** caused by the bump — the same hang happened on 2026-08-04 on an unrelated dependabot branch — but it blocks this PR from going green, so it is fixed here rather than parked.

`Start()` waited for a previous command still in `STOPPING` while holding `cmdLock`, the same lock `IsRunning()`, `LastExitCode()` and `Stopped()` take. A process that never reaches a final state — one ignoring `SIGTERM`, or whose children keep its stdout/stderr open so `exec.Cmd.Wait` never returns — left that lock held forever and wedged the whole superviser.

- `runLock` now serializes the `Start`/`Stop` lifecycle and may be held across waits; `cmdLock` only guards the `cmd` field and is never held while waiting.
- `Start()` gives up after `stoppingWaitTimeout` (30s) with an error instead of blocking indefinitely.
- A failed `cmd.Stop()` deliberately keeps the command in place — dropping it would let a later `Start()` spawn a second process over the same data.
- overseer state is read through its locked accessors; `Cmd.State` is guarded by an unexported lock and reading it directly raced with the command's own goroutine.
- The tests fed log lines into an unbuffered channel, so the read loop blocked, stopped draining overseer's stdout/stderr and kept the command from ever completing. That is what made the deadlock reachable from CI.

## Verification

- New regression test asserts the observers stay responsive while `Start()` waits; confirmed it fails against the old locking and passes against the new.
- `go test ./...` and `go test -race ./node-manager/superviser/` pass, plus a stress run of 4 concurrent instances × 25 iterations on a CPU-throttled Linux container shaped like CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you s
 
 - The `firehose` app now restarts when its block hub can no longer link incoming live blocks, instead of hanging every request at a frozen head indefinitely. A live-source gap whose one-block files were already merged away can never be linked, and `head_block_number`/`finalized_block_number` keep tracking the live source, so the process looked healthy throughout.
 
+- Restarting a supervised node process no longer deadlocks the whole superviser. `Start` waited for a previous process still stopping while holding the lock that also guards `IsRunning`, `LastExitCode` and `Stopped`, so a process that never reached a final state — one ignoring `SIGTERM`, or whose children keep its stdout/stderr open — froze every one of those calls forever. Waiting no longer holds that lock, and `Start` now gives up after 30s with an error instead of blocking indefinitely.
+
 ## v1.17.0
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,13 +14,15 @@ If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you s
 
 - Bumped `firehose-networks` to [v0.2.3](https://github.com/streamingfast/firehose-networks/releases/tag/v0.2.3): Added Ethereum Hoodi testnet (`hoodi`, `eip155:560048`)  StreamingFast Firehose and Substreams endpoints (`hoodi.eth.streamingfast.io:443`).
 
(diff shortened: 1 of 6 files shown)


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/firehose-core:ad92882-amd64</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/firehose-core:ad92882-amd64</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:3f6b8731a722776661879a7804d2a6971c6a79e55a002868ef9da5d693810ede</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
<tr><td>size</td><td>164 MB</td></tr>
</table>
</details>
<details><summary>:package: Base Image</strong> <code>oisupport/staging-amd64:24.04</code></summary>
<table>
<tr><td>also known as</td>
<td><ul>
<li><code>a215e986b44aae6f10795ded1e39ce93d9c236d8163d21a522ffd0ab3659f546</code></li>

<li><code>noble</code></li>

<li><code>noble-20260730.1</code></li>
</ul>
</td></tr>
<tr><td>digest</td><td><code>sha256:019e8eb29a85e74d64925745884f2ec79aa27e3feab36353d24656f4d6b89467</code></td></tr>
<tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 5 low: 4 </td></tr>
</table>
</details>

<table></table>

PR #218, opened 2026-08-20
Handle empty merged-blocks bundles, fix clock-only client stall

sduchesneau:
Found while investigating stalled Firehose requests over a Solana range with skipped block numbers, on a store holding an empty `0096336100.dbin.zst` (blocks jump from `96336011` to `96336220`).

Three unrelated bugs came out of it, plus one test-only addition.

## 1. `tools firehose-client --print-clock-only` hangs at end of stream

`cmd/tools/firehose/client.go`. The goroutine draining `resps` is only started on the block-decoding path, but only `printCursorOnly` returned early — the clock path fell through to `<-allDone`, which nothing ever closes. Every block printed, then the command sat there forever. `--print-clock-only` now returns early too.

This is why `firecore tools firehose-client --print-clock-only <ep> 300:400` looked like a server stall: the server had already answered `InvalidArgument`.

## 2. Merger crashes at startup when the most recent bundles hold no block

`merger/merger_io.go`. On chains that skip block numbers a whole bundle range can contain nothing, and the merger writes a merged-blocks file with a DBIN header and no block so boundaries stay contiguous (`Bundler`'s skip-forward loop → `MergeAndStore` filtering every carried block out).

`NextBundle` reads the last block back out of the last file to publish the LIB. For such a file `lastBlock()` returns `nil, nil`, and `readLastBlockFromMerged` then dereferenced it:

```go
t := last.Timestamp.AsTime()   // nil *pbbstream.Block
```

Startup now walks back to the last bundle that actually holds a block, bounded by `lowestBaseBlock`. A file without even a DBIN header stays an error — that is a broken file, not an empty bundle.

## 3. Bumped `bstream` — unreadable merged-blocks file stalls the stream

streamingfast/bstream#65. `FileSource`'s per-file reader goroutine calls `Shutdown()` on failure without ever closing that file's blocks channel, and the consumer was a plain `range` over it. Whichever won the race decided whether the request returned the error or hung forever. That was the original stall.

A zero-byte merged-blocks file (no DBIN header, written by mergers predating the streaming bundle writer) still fails, now deterministically:
`processing of file "0096336100" failed: unable to create block reader: unable to read file header: EOF`

## 4. Test only: bundler behaviour across skipped bundle ranges

`TestBundlerSkipsWholeBundles` drives the real `DStoreIO` with a chain skipping one and then four full bundle ranges, and asserts what actually lands in the store:

| bundle | blocks |
|---|---|
| `0000000100` | 100–106 |
| `0000000200` | — (header only) |
| `0000000300` | 320, 321, 322 |
| `0000000400` | 450, 451 |
| `0000000500`–`0000000800` | — (header only) |

Every boundary gets a file, none is skipped, and no block lands outside the bundle its number belongs to. No production code changed for this one — it pins behaviour that both fixes above depend on.

## Verified

Against a Solana merged-blocks store with the gap, no live source, `96336000:96336300`, 3 runs each:

- zero-byte `0096336100` → fails fast with the header error, every time (used to hang about half the time)
- header-only `0096336100` → streams 53 blocks across the gap, clean `stop block reached`

`--print-clock-only` returns instead of hanging in both cases.


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/firehose-core:0bae620-amd64</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/firehose-core:0bae620-amd64</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:5a59f927e03bd965dc5717accb9cd766365235f2d7ee632739a9cfcd6c7d55b7</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
<tr><td>size</td><td>160 MB</td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>

<li><code>noble</code></li>

</ul>
</td></tr>
<tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 3 low: 4 </td></tr>
</table>
</details>

<table></table>

billettc (approved):

PR #220, opened 2026-08-20
fix(tools): check merged-blocks detects blocks beyond their bundle boundary

YaroShkvorets:
`tools check merged-blocks` only listed file names, so a "bloated" bundle holding a block from a later range left no hole and passed as `🆗 No hole found`. Running it with `-s` also didn't catch this problem, while every Firehose/Substreams read over it failed with:

```
merged blocks file "0096336100" contains block 96336220, beyond the configured
bundle size of 100 blocks
```

- New `-b/--validate-blocks`: reads each file and checks every block against the bundle its name claims — same rule `bstream`'s `FileSource` enforces at serving time. `-s`/`-f` imply it.
- Boundaries are checked before `--range` filtering, so an offending block can't be skipped as out-of-range.
- Blocks *below* the base num stay accepted: the merger carries the previous bundle's last block over (`merger/bundler.go`), readers skip them.
- Also reports out-of-order blocks; the "contained only N blocks" line no longer hardcodes 100.
- Without the flag the summary says the content was not validated instead of implying integrity was checked.
- `CheckMergedBlocks` keeps its signature (deprecated wrapper); real entry point is `CheckMergedBlocksWithOptions`.

Output on a bloated bundle:

```
❌ Merged blocks file 0096336100 holds 1 block(s) beyond its bundle boundary [#96 336 100, #96 336 199] (#96336220), it will fail any read configured with a bundle size of 100

Summary:
> 🆗 No hole found
> 🆘 1 merged-blocks file(s) hold blocks beyond their own bundle boundaries: [0096336100]
```

Tests: unit tests on the validator plus two end-to-end runs against a local dstore (holes-only pass stays silent, `-b` pass names the file).


CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you s
 
 ### Fixed
 
+- `firecore tools check merged-blocks` no longer reports a range as healthy when a merged-blocks file holds blocks that do not belong to it. It only listed the store and looked for holes in the sequence of file names, so a bundle such as `0096336100` holding block `96336220` passed with "🆗 No hole found" while every Firehose and Substreams read over that range failed with `merged blocks file "0096336100" contains block 96336220, beyond the configured bundle size of 100 blocks`. The new `--validate-blocks` (`-b`) flag reads each file and checks every block against the bundle boundaries claimed by its file name, the same rule `bstream`'s `FileSource` enforces at serving time, and reports the offending files in the summary. `--print-stats` and `--print-full` imply it. Without the flag the check stays a file-name listing, and the summary now says so instead of implying the content was verified.
```
sduchesneau: >no longer reports a range as healthy
this is misleading.  it doesn't report it as unhealthy now. That's not the change. The change is the new flag.
Description too LLMesque, suggested:

```
Added flag `--validate-blocks` (`-b`) on `firecore tools check merged-blocks`, ensures that individual block sequence is checked (reading the content of the merged blocks, instead of just checking by file name). Flag is implied by `--print-stats` and `--print-full`
```

cmd/tools/check/blocks.go
```diff
@@ -147,52 +174,177 @@ func CheckMergedBlocks[B firecore.Block](ctx context.Context, chain *firecore.Ch
 		fmt.Printf("> 🆗 No hole found\n")
 	}
 
+	switch {
+	case len(brokenSegments) > 0:
+		fmt.Printf("> 🆘 %d merged-blocks file(s) hold blocks beyond their own bundle boundaries: %s\n", len(brokenSegments), formatBrokenSegments(brokenSegments))
```
sduchesneau: Misleading error, try something like:
```
fmt.Printf("> 🆘 %d merged-blocks file(s) are either unreadable or hold blocks beyond their own bundle boundaries: %s\n", len(brokenSegments), formatBrokenSegments(brokenSegments))
fmt.Println(">    See the ❌ lines above for which is which. Any Firehose or Substreams request covering them fails. You can try `tools fix-bloated-merged-blocks` on the latter.")
```

YaroShkvorets:
addressed pr comments

sduchesneau (approved):

PR #222, opened 2026-08-21
Normalize since/date-range parsing across substreams logs commands

maoueh:
## Summary

`firecore tools substreams logs connection` (singular) and `logs connections` (plural) handled time ranges through two unrelated interfaces — a positional `[<date-range>]` argument on one, a `--since`/`--date-range` flag pair on the other — with incompatible parsers underneath. Copying a value between the two broke, most visibly `--since 2d` on `connections`, which Go's native `time.Duration` flag parser rejects outright (no `d` unit).

- `logs connection` now takes `--since`/`--date-range` flags instead of a positional date-range argument, matching `logs connections`.
- Both commands share the same parsing: `--since` is a relative duration (`30m`, `2h`, `1d`, `1w`, `"1 day ago"`, …), `--date-range` accepts slash, colon, or bare-timestamp forms.
- All three commands (`connection`, `connections`, `reexec`) gained a `1w` shortcut. Units are fixed-length to dodge calendar ambiguity: `d` == 24h, `w` == 7d == 168h.
- `logs reexec` keeps its positional date-range argument (out of scope here) but shares the same underlying relative-duration parser, so `1w` works there too.

Net effect: a since/date-range value copied from a `logs connections` listing straight into `logs connection <trace-id> --since ...`/`--date-range ...` now works unchanged — the workflow this was built around.

## Trade-offs

- `logs connection`'s date-range is no longer a positional argument — this is a breaking CLI change for that command, done deliberately to normalize with `logs connections` rather than keep two shapes.
- Dropped the undocumented "trailing slash means open-ended end" form (`--date-range "2024-01-15T10:00:00Z/"`) — a bare timestamp without any separator already means the same thing.

## Test plan
- [x] `go build ./...`
- [x] `go test ./cmd/...`
- [x] Manually exercised `logs connection <trace-id> --since 1w`, `logs connections --since garbage` (clean error), `--date-range` with colon separator via `go run ./cmd/firecore tools substreams logs ...`


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/firehose-core:e8aa16f-amd64</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/firehose-core:e8aa16f-amd64</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:6127594a5d34a42141ea00d68603921f186d6e7db7fe917cd00e6daf696c3d65</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
<tr><td>size</td><td>172 MB</td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>

<li><code>noble</code></li>

</ul>
</td></tr>
<tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 24 low: 9 </td></tr>
</table>
</details>

<table></table>

cmd/tools/substreams/logs.go
```diff
@@ -62,8 +69,8 @@ Note: Currently only GCP Cloud Logging backend is supported.`,
 	}
 
 	cmd.Flags().String("backend", "gcp", "Log backend to use (currently only 'gcp' supported)")
-	cmd.Flags().Duration("since", time.Duration(0), "Look back duration (e.g., '1h', '30m', '2d'). Mutually exclusive with --date-range")
-	cmd.Flags().String("date-range", "", "Date range in format '<start>[/<end>]'. End defaults to now. Mutually exclusive with --since")
+	cmd.Flags().String("since", "", sinceFlagHelp)
```
sduchesneau, line 69: That's weird with those two flags possibly doing the same thing..

functionally identical:
```
--since=1h
--date-range=1h
```

maybe we should have just `--since=` and `--until=` ?
like:

`--since=24h` and `--until=1h`
`--since=2026-05-21T10:23:12Z` and `--until=2026-05-22T10:23:12Z`

We could also accept `--since=24h` and `--until=+2h30m`

So there wouldn't be overlap between the function of both flags.

Either you use just `--since` with a duration (most common case for "last x hours")
Either you use both with durations `--since=48h` `--until=24h` (sous-entendu: "ago")
Either you use timestamps
Either you use timestamp for since, and use an `until` that is relative to the `since` (with `+` prefix)

maoueh: Keeping just --since then, let's keep it simple

PR #225, opened 2026-08-25
Add substreams prune-states and prune-outputs tools

sduchesneau:
Companion of streamingfast/substreams#901 (tier1 resuming from the last remaining store snapshot). Three things:

## `firecore tools substreams prune-states <store-url> --keep-every <blocks> --truncate-below-block <block>`

Deletes intermediate store snapshots (`<end>-<start>.kv[.zst]` under each module's `states/`), keeping the first snapshot of each `--keep-every`-aligned window (windows start on multiples of `--keep-every` whatever the module's initial block is, so misaligned snapshotting cadences don't lose everything). Only snapshots whose end block is at or below `--truncate-below-block` are thinned: everything above that block is kept untouched, so requests in flight and the live head keep their nearby resume points, and the most recent snapshot of each module is always kept. `--minimum-age` (e.g. `3d`), if set, additionally spares any snapshot last modified more recently than that — modification times come straight out of the dstore listing, nothing is downloaded. Partials, outputs and index files are untouched. `--dry-run`, `--parallelism`.

Module folders are discovered the same way `purge` does — direct tier1 layouts (`<store-url>/<hash>`, `<store-url>/<tag>/<hash>`) and shared network roots (`<store-url>/<network>/substreams-states/<tag>/<hash>`) — and only each module's `states/` prefix is then listed, so the scan never enumerates the output files next to the snapshots. A URL pointing inside a single module folder (or at its `states/` folder) falls back to walking the whole tree. Works on gs/s3/azure/file through dstore.

Not a replacement for the `purge` tool: purge removes whole unused modules, prune thins the caches of modules still in use.

## `firecore tools substreams prune-outputs <store-url> --network <net> --truncate-below-block <block> --minimum-age <age>`

Deletes execution output files (`<start>-<end>.output[.zst]` under each module's `outputs/`) that are BOTH fully below `--truncate-below-block` AND last modified longer than `--minimum-age` ago (e.g. `3d`). Modification times come straight out of the dstore listing, nothing is downloaded. Nothing outside `outputs/` folders is ever touched. Same purge-style module discovery, cache tags included.

Module folders carrying an spkg (`substreams.spkg.zst` / `substreams.partial.spkg.zst`) were requested directly as output modules, not as intermediate dependencies, so their outputs are what serves those requests: they are skipped by default. `--output-module-minimum-age`, if set, prunes them too using that (typically longer) age. `--dry-run` (`-n`), `--force`, `--parallelism`.

## Substreams bump

Bumps `substreams` to `v1.22.1-0.20260826153554-edb28d2ecdae`: prunable fullKV snapshots (tier1 walks back to the last block where every store still has one), `FailedPrecondition` instead of an infinite retry when a snapshot a running request needs is deleted, re-produced mapper outputs, bounded initial store lookup, progress-message cadence ramp, and per-plan `last_used_<plan>` markers (what `purge --retention plan=duration` reads).

## Tests

The bench that exercised this (scenarios + fuzzer over pruned/deleted caches) lives in substreams as the opt-in `tests_e2e` suite `TestThinstore` (streamingfast/substreams#901), not here. Prune selection logic, filename/spkg parsing and discovery are unit-tested in this repo.


sduchesneau:
<h2>:mag: Vulnerabilities of <code>ghcr.io/streamingfast/firehose-core:41aa669-amd64</code></h2>

<details open="true"><summary>:package: Image Reference</strong> <code>ghcr.io/streamingfast/firehose-core:41aa669-amd64</code></summary>
<table>
<tr><td>digest</td><td><code>sha256:73f31704dd88df493d8247f6eb0ef49614674180b1f0b44321b8da81f894d4ee</code></td><tr><tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 0 low: 0 </td></tr>
<tr><td>size</td><td>164 MB</td></tr>
</table>
</details>
<table>
<tr><td>also known as</td>
<td><ul>

<li><code>noble</code></li>

</ul>
</td></tr>
<tr><td>vulnerabilities</td><td>critical: 0 high: 0 medium: 8 low: 7 </td></tr>
</table>
</details>

<table></table>

devel/thinstore/.gitignore
```diff
@@ -0,0 +1,5 @@
+firehose-data/
```
maoueh: I don't like this overall thinstore setup. 

Want more Go and to be integrate in our Go test suite, Golang support a Fuzzer, seems like good tests we should have running on CI.

sduchesneau: I agree! I didn't think Claude would push this as a PR honestly, it was just local fuzzing experiments 😓

sduchesneau: All e2e tests moved to https://github.com/streamingfast/substreams/pull/901 under tests_e2e (gated with env var because they take like 20 minutes++)

cmd/tools/substreams/prune_states.go
```diff
@@ -0,0 +1,337 @@
...
+			continue
+		}
+		out = append(out, file)
+	}
+	return out
+}
+
+func deleteAll(ctx context.Context, store dstore.Store, files []string, parallelism int) error {
+	group, groupCtx := errgroup.WithContext(ctx)
+	group.SetLimit(parallelism)
```
maoueh, line 342: This parallelism and the one on listModulesSnapshots shared the same config, so technically one could have 2x the operations at one single point.

I think it's a big problem in this context, the default value could be scaled down or not.

maoueh (approved):

cmd/tools/substreams/prune_states.go
sduchesneau: I double checked, the two phases are sequential so no issue there \o/

PR #229, opened 2026-09-01
Refresh Cosmos well-known proto and add compare-blocks --fields

billettc:
Well-known Cosmos descriptors are regenerated from Buf (commit 3c05174). The block model now includes CometBFT v1 consensus params (abci, synchrony, feature), a bls12381 public key, and a local Int64Value for feature heights, so those fields are no longer unknown when comparing Injective blocks. The generator discards unknown Buf descriptor extensions so go generate works against current BSR JSON.

compare-blocks --fields lists the protobuf paths that actually differ when --diff prints nothing because proto.Equal includes unknown fields while the JSON diff strips them.


sduchesneau (approved):

sduchesneau:
@copilot resolve the merge conflicts in this pull request

PR #230, opened 2026-09-01
Record block count, size and first block time on merged blocks

sduchesneau:
Records what a merged-blocks file holds on the object itself, as three custom metadata entries, so a listing tells you without reading the file:

| key | value |
| --- | --- |
| `datasize` | the file's size once decompressed, in bytes |
| `itemcount` | the number of blocks it holds |
| `timestamp` | the time of its first block, `2025-10-12 10:23:12` in UTC |

Google Cloud Storage only, everywhere. No other backend keeps custom object metadata a listing reads back, so elsewhere nothing is written and nothing changes.

### The merger writes them as it uploads

The bundle is streamed, so its uncompressed size is counted on the way through rather than by reading the file back. The first block's time is read from the head of the first one-block file, far enough to reach the timestamp field and no further, so a chain with large blocks costs no more than one with small blocks.

An annotation that cannot be written never fails the merge: the bundle is written and correct, only its description is missing, and the tool below fills it back in.

### `firecore tools annotate-merged-blocks <gs-store-url>`

Backfills the same three entries on files written before the merger did it.

No block is ever held in memory: each one is read into a scratch buffer reused from block to block and from file to file, and dropped; of the first block only the head is looked at, far enough to reach its timestamp. Only the block's metadata fields are decoded, its chain-specific payload being skipped, so no chain block type is needed. `--parallelism` (32 by default) sets how many files are read at once.

Files already carrying all three entries are skipped straight from the listing, which carries their metadata, so a rerun over a mostly-done store costs one listing and nothing else — `--overwrite` recomputes them. `--start-block` and `--stop-block` bound the listing service-side, `--dry-run` reads nothing.

Reads and metadata writes are both pinned to the generation and metageneration the listing returned, so a file the merger rewrites mid-run is reported as failed rather than annotated with values from a version that is gone.

`--grpc` switches to the Cloud Storage gRPC API, which on GKE takes the DirectPath route straight to the storage backend, spread over `--grpc-connection-pool` connections (one per 32 of `--parallelism` by default). The store URL takes the same two query parameters `dstore` reads: `?project=` bills the reads to that project on a requester-pays bucket, and `?client_protocol=grpc` selects the same transport as `--grpc`. Setting a project turns DirectPath off, since DirectPath does not carry the `x-goog-user-project` header requester-pays billing needs.

### `firecore tools stats-merged-blocks <gs-store-url>`

Reports total compressed and uncompressed size, block count, compression ratio and bytes per block, broken down by month and totalled over `--start-block` to `--stop-block`.

Nothing is downloaded — every number comes from the three entries above, which come back with the listing, so the whole report costs one listing however large the range is. Files missing any of the three are counted and reported, and contribute to nothing.

```
MONTH     FILES    BLOCKS  COMPRESSED  UNCOMPRESSED  RATIO  COMP./BLOCK  UNCOMP./BLOCK
2025-08   1,000   100,000      1.2 GB        4.5 GB  3.75x        12 kB          45 kB
2025-09   1,000   100,000      1.3 GB        4.9 GB  3.77x        13 kB          49 kB

TOTAL     2,000   200,000      2.5 GB        9.4 GB  3.76x        12 kB          47 kB
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,14 @@ If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you s
 
 ### Added
 
+- Added `firecore tools annotate-merged-blocks <gs-store-url>` which records the uncompressed size of every merged-blocks file as a `datasize` custom metadata entry on the object, so the size can be read from a listing instead of by downloading the file.
+
+  Google Cloud Storage only, since it writes object metadata without rewriting the file. Each file is streamed through a zstd decoder into a byte counter and dropped as it decodes, so memory stays flat whatever the file size, and `--parallelism` (32 by default) sets how many files are read at once. Files that a previous run already annotated are skipped straight from the listing, which carries their metadata, so a rerun over a mostly-done store costs one listing and nothing else; `--overwrite` recomputes them. A plain `.dbin` file is never read at all, its object size is already the answer. `--start-block` and `--stop-block` bound the listing service-side, `--dry-run` reports what would be annotated without reading anything.
+
+  Reads and metadata writes are both pinned to the generation and metageneration the listing returned, so a file the merger rewrites mid-run is reported as failed rather than annotated with the size of a version that is gone. `--grpc` switches to the Cloud Storage gRPC API, which on GKE takes the DirectPath route straight to the storage backend, spread over `--grpc-connection-pool` connections (one per 32 of `--parallelism` by default).
+
+  The store URL takes the same two query parameters `dstore` reads off a `gs://` URL: `?project=` bills the reads to that project on a requester-pays bucket, and `?client_protocol=grpc` selects the same transport as `--grpc`. Setting a project turns DirectPath off, since DirectPath does not carry the `x-goog-user-project` header requester-pays billing needs.
+
 - `firecore tools substreams purge`, `prune-states` and `prune-outputs` all skip any module folder carrying a `DO_NOT_PRUNE` file at its root, next to the `last_used*.zst` markers, and report how many folders that spared.
 
 - Added `firecore tools last-oneblock <oneblocks-store>` which prints the highest block number found among the store's one-block files, as a bare number on stdout, exiting non-zero when the store cannot be listed, holds no one-block file, or a filename does not parse as one.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,13 +16,23 @@ If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you s
 
 ### Added
 
-- Added `firecore tools annotate-merged-blocks <gs-store-url>` which records the uncompressed size of every merged-blocks file as a `datasize` custom metadata entry on the object, so the size can be read from a listing instead of by downloading the file.
+- The merger now records what a merged-blocks file holds on the object itself, as three custom metadata entries written as it uploads each bundle: `datasize`, the file's size once decompressed, `itemcount`, the number of blocks it holds, and `timestamp`, the time of its first block written as `2025-10-12 10:23:12` in UTC. A listing then tells what a file holds without reading it.
 
-  Google Cloud Storage only, since it writes object metadata without rewriting the file. Each file is streamed through a zstd decoder into a byte counter and dropped as it decodes, so memory stays flat whatever the file size, and `--parallelism` (32 by default) sets how many files are read at once. Files that a previous run already annotated are skipped straight from the listing, which carries their metadata, so a rerun over a mostly-done store costs one listing and nothing else; `--overwrite` recomputes them. A plain `.dbin` file is never read at all, its object size is already the answer. `--start-block` and `--stop-block` bound the listing service-side, `--dry-run` reports what would be annotated without reading anything.
+  Google Cloud Storage only: no other backend keeps custom object metadata a listing reads back, so elsewhere nothing is written and nothing changes. The size is counted as the bundle streams to the store, and the first block's time is read from the head of the first one-block file, far enough to reach the timestamp field and no further, so a chain with large blocks costs no more than one with small blocks. An annotation that cannot be written never fails the merge: the bundle is written and correct, only its description is missing, and `firecore tools annotate-merged-blocks` fills that back in.
 
-  Reads and metadata writes are both pinned to the generation and metageneration the listing returned, so a file the merger rewrites mid-run is reported as failed rather than annotated with the size of a version that is gone. `--grpc` switches to the Cloud Storage gRPC API, which on GKE takes the DirectPath route straight to the storage backend, spread over `--grpc-connection-pool` connections (one per 32 of `--parallelism` by default).
+- Added `firecore tools annotate-merged-blocks <gs-store-url>`, which writes those same three entries on merged-blocks files written before the merger did it. Google Cloud Storage only, for the same reason.
+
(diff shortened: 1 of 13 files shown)


cmd/tools/mergeblock/merged_blocks_gs.go
```diff
@@ -0,0 +1,171 @@
...
+// gsTarget is what the store URL says: where to look, plus the two query parameters dstore
+// reads off a 'gs://' URL, so the same URL means the same thing here as it does anywhere else
+// in the stack.
+type gsTarget struct {
+	bucket string
+	prefix string
+	// userProject bills the reads to that project, for a requester-pays bucket ('?project=').
+	userProject string
+	// grpc is '?client_protocol=grpc', which selects the same transport as --grpc.
+	grpc bool
```
maoueh, line 42: Why it didn't use `dstore` here? Even if just for GS, not sure it's worth not using dstore?

cmd/tools/mergeblock/merged_blocks_gs.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	pool := (parallelism + 31) / 32
+	if pool < 1 {
+		return 1
+	}
+	if pool > 16 {
+		return 16
+	}
+	return pool
+}
```
maoueh, line 104: Interesting, would be useful in dstore?

cmd/tools/mergeblock/tools_annotate_merged_blocks.go
```diff
@@ -0,0 +1,448 @@
...
+			exclusive, so setting a project turns DirectPath off and routes the gRPC traffic
+			through the Google Front End, which is the only side that enforces the billing.
+		`),
+		Example: cli.Dedent(`
+			# Annotate a whole merged-blocks store with 64 concurrent readers
+			firecore tools annotate-merged-blocks gs://example-bucket/eth-mainnet/merged --parallelism 64
+
+			# Annotate blocks 1000000 to 2000000 over gRPC, from a GKE pod
+			firecore tools annotate-merged-blocks gs://example-bucket/eth-mainnet/merged \
+			  --start-block 1000000 --stop-block 2000000 --grpc
```
maoueh, line 81: We are confident enough we don't need a resume operation?

cmd/tools/mergeblock/tools_annotate_merged_blocks.go
```diff
@@ -0,0 +1,448 @@
...
+			chunk = length
+		}
+		read, err := io.ReadFull(source, scratch[:chunk])
+		length -= int64(read)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
maoueh, line 410: Isn't there a io. helper for doing this? Maybe not for a fixed length of discard.

merger/block_time.go
```diff
@@ -36,80 +36,14 @@ func readBlockTimestamp(ctx context.Context, obf *bstream.OneBlockFile, opener f
...
 	if _, err := io.ReadFull(r, buf); err != nil {
 		return time.Time{}, fmt.Errorf("reading block message bytes: %w", err)
 	}
 
-	return extractTimestampFromProto(buf)
-}
-
-// extractTimestampFromProto scans a partial pbbstream.Block protobuf payload
-// field by field, returning as soon as field 4 (google.protobuf.Timestamp) is found.
-func extractTimestampFromProto(data []byte) (time.Time, error) {
```
maoueh, line 56: That is wild ahah :)

CHANGELOG.md
```diff
@@ -16,6 +16,24 @@ If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you s
 
 ### Added
 
+- The merger now records what a merged-blocks file holds on the object itself, as three custom metadata entries written as it uploads each bundle: `datasize`, the file's size once decompressed, `itemcount`, the number of blocks it holds, and `timestamp`, the time of its first block written as `2025-10-12 10:23:12` in UTC. A listing then tells what a file holds without reading it.
+
+  Google Cloud Storage only: no other backend keeps custom object metadata a listing reads back, so elsewhere nothing is written and nothing changes. The size is counted as the bundle streams to the store, and the first block's time is read from the head of the first one-block file, far enough to reach the timestamp field and no further, so a chain with large blocks costs no more than one with small blocks. An annotation that cannot be written never fails the merge: the bundle is written and correct, only its description is missing, and `firecore tools annotate-merged-blocks` fills that back in.
```
maoueh, line 21: While the `annotate-megred-blocks` might find that too prohibitive, on the merger S3 should be supported no?

maoueh (changes requested):
Questions mostly

CHANGELOG.md
sduchesneau: we don't know the full size or objectcount beforehand (because of the optimisations for direct writing and memory reduction), so we would have to write the file, THEN do a copy operation ($ + time) again after writing a merged block. Sounds prohibitive to me.

cmd/tools/mergeblock/tools_annotate_merged_blocks.go
sduchesneau: it's pretty cheap to run twice: on gcs, the LISTING can include the metadata, so you list and skip anything that already contains all 3 fields so you don't have to recompute it.
So if you get interrupted, run it again ;)

cmd/tools/mergeblock/merged_blocks_gs.go
sduchesneau: hmmm maybe! I'll see if I can PR in dstore

cmd/tools/mergeblock/merged_blocks_gs.go
sduchesneau: Optimized for max performance... missing a few things in dstore and I didn't want to bloat it too much !

4 reasons:

1. Skip already-annotated files from the listing alone — needs Metadata on each listed object, so isAnnotated costs no request: missing Metadata map[string]string on ObjectEntry and "Metadata" in WalkAttributes' SetAttrSelection (gsstore.go:391). S3's ListObjectsV2 can't return it, so S3Store falls back to a per-object fetch.

2. Bound the listing to --start-block/--stop-block — needs the bounds pushed down as StartOffset/EndOffset rather than listing the store and filtering: missing WalkAttributesFromTo() (WalkFromTo exists for names only).

3. Reuse one zstd decoder per worker instead of one per file — uncompressedReader (common.go:186) calls zstd.NewReader(reader) per object with default options (GOMAXPROCS decode goroutines, full-size window), against 32 decoders total with WithDecoderConcurrency(1) + WithDecoderLowmem(true) and a Reset per file: missing decoder reuse in commonStore, best as a sync.Pool of *zstd.Decoder returned on Close — no API change, and it fixes every dstore caller. A dstore.WithZstdDecoderOptions(...) option would bound each decoder's cost but still allocate one per file.

4. Spread concurrent readers over several gRPC connections — one HTTP/2 connection multiplexes every stream over one socket and caps throughput below 32 workers; the storage client's defaultConnPoolSize is 1 (grpc_client.go:56): missing option.WithGRPCConnectionPool plumbing, as a ?connection_pool=N URL parameter or a dstore.WithGRPCConnectionPool(n) option.

Want me to add those to dstore ? I know 4. would be a good addition, but the other ones may be starting to add complexity..
... since we only support gcs, I thought this would be a good approach... let me know @maoueh !

cmd/tools/mergeblock/tools_annotate_merged_blocks.go
sduchesneau: true, pushing simplfied code

cmd/tools/mergeblock/merged_blocks_gs.go
maoueh: Nope let's keep out for now, do something if it feels important to you only.

CHANGELOG.md
maoueh: Ok I understand, we would actually need the uncompress size for each one block directly, then we could do that inflight computation.

Good enough for now then, if somehow wants this, they can work on it.

maoueh (approved):

PR #231, opened 2026-09-01
fix(blockpoller): only warn when client calls are actually failing

YaroShkvorets:
On slower chains like Bitcoin/Litecoin, poller logs this warning every 15 seconds:
```
WARN (firebtc.poller) no clients have been working for over 1 minute, still retrying 
```
Basically it's treating lack of new blocks as failing RPC calls.

This PR changes the warning logic to warn only if calls are actually failing for over a minute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you s
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


sduchesneau (approved):
LGTM. First I thought we wanted to keep it to know that upstream is not progressing, but that will be seen indrectly through the other (drift) metrics, so that log works better your way.

sduchesneau:
@YaroShkvorets please make branch up-to-date so I can merge

PR #234, opened 2026-09-02
prometheus ready metric shows 1 when firehose is blocked on not-available first streamable block

matthewdarwin:
prometheus ready metric shows 1 when firehose is blocked on not-available first streamable block.  It needs to not be ready until it can get the first block.

firehose is not listening on http connections until it can read the first block, so clearly it is not ready.


GabrielCartier:
The move out of the `forkableHub.Ready` select is the right fix, but it stops one line short: on the failure path the goroutine keeps going and sets the metric anyway.

```go
if err := a.modules.InfoServer.Init(ctx, forkableHub, mergedBlocksStore, oneBlocksStore, a.logger); err != nil {
    a.Shutdown(fmt.Errorf("cannot initialize info server: %w", err))
}
// falls through
a.isReady.CAS(false, true)
if withLive { metrics.AppReadiness.SetReady() }
```

`a.Shutdown` is asynchronous, so nothing stops the goroutine. In exactly the scenario this PR is about — first streamable block not available — `Init` blocks until the 3 minute context timeout, returns the ctx error, and `ready` goes to 1 regardless. `isReady` gets set too; `/healthz` only escapes it because `IsReady` checks `IsTerminating`. A `return` after `a.Shutdown` covers both, and then the CHANGELOG line ("only changes to 1 after initialization succeeds") is accurate.

Separate question, not a blocker: the `withLive` guard preserves today's behaviour, but it means a merged-blocks-only firehose reports `ready=0` forever while serving fine. That was already true before this PR, but the CHANGELOG now states it as intent — worth confirming that is what you want rather than something inherited.

GabrielCartier (changes requested):
check comments

matthewdarwin:
hm wasn't thinking of running firehose without having a relayer.  Does anyone actually do that in production?  For local development you're probably not checking the app ready status anyway.

sduchesneau:
@matthewdarwin I still think you should just remove the `if withLive` and that would be good for me

matthewdarwin:
@sduchesneau @GabrielCartier copilot took a pass at addressing your comments.

maoueh (approved):

PR #237, opened 2026-09-03
Add request instrumentation to Firehose Blocks logs

maoueh:
- Emit an "incoming firehose Blocks request" log line when a request starts (`trace_id`, `organization_id`, `api_key_id`, `real_ip`, `start_block`, `stop_block`, `final_blocks_only`, `cursor`), mirroring Substreams' incoming/stats two-line design so both can be consumed by the same analytics model.
- Enrich the existing "firehose process completed" line with `duration`, `time_to_first_data`, and `resolved_start_block`; message text and existing fields are unchanged.
- Both lines now fire on every request path — early rejections (session denied, rate limited, unimplemented transforms) and client disconnects included — via a deferred completion log, instead of only on the happy path.
- Normalize the completion line's `error` field to the literal `"context canceled"` on client cancellation, so it's distinguishable from real failures downstream.
- Update CHANGELOG.md.


firehose/server/blocks.go
```diff
@@ -79,18 +79,77 @@ func (s *Server) Block(ctx context.Context, request *connect.Request[pbfirehose.
...
+	)
+
+	defer func() {
+		logErr := err
+		if streamRan {
+			logErr = runErr
+		}
+		if errors.Is(logErr, context.Canceled) {
+			logErr = context.Canceled
+		}
```
GabrielCartier, line 131: This is the blocker. `context.WithCancelCause` at the top of the function means `ctx.Err()` is plain `context.Canceled` no matter *why* the context died, so this collapses server-initiated cancellation into the client-disconnect bucket.

Concretely: `s.sessionPool.Get`'s revocation callback fires mid-stream with `dsession.ErrQuotaExceeded` → `cancelRunning(connectErr)` → `str.Run` returns `context.Canceled` → this branch rewrites it to the sentinel. The client correctly receives the `*connect.Error` cause (line 374-375), but the completion line reads `error: "context canceled"`. Quota-exceeded terminations become indistinguishable from someone closing their laptop, which is the exact distinction the PR body says this normalization creates.

`context.Cause(ctx)` is available right here and already used 250 lines down. Something like:

```go
logErr := err
if streamRan {
    logErr = runErr
}
if errors.Is(logErr, context.Canceled) {
    if cause := context.Cause(ctx); cause != nil && cause != context.Canceled {
        logErr = cause
    } else {
        logErr = context.Canceled
    }
}
```

Related: because `errors.Is` matches wrapped errors, this also discards the wrapper on any error whose chain contains `context.Canceled`. If the filesource returns `fmt.Errorf("reading merged blocks %s: %w", file, ctx.Err())`, the file identity is gone from the completion line, and the Debug at line 372 only fires when `ctx.Err() != context.Canceled`, so it's lost from both.

firehose/server/blocks.go
```diff
@@ -79,18 +79,77 @@ func (s *Server) Block(ctx context.Context, request *connect.Request[pbfirehose.
...
+		if firstDataSent {
+			fields = append(fields,
+				zap.Duration("time_to_first_data", timeToFirstData),
+				zap.Uint64("resolved_start_block", resolvedStartBlock),
+			)
+		}
+		if auth != nil {
+			fields = append(fields,
+				zap.String("api_key_id", apiKeyID),
+				zap.String("user_id", organizationID),
```
GabrielCartier: `user_id` changes meaning here. It was `auth.UserID()` before this PR; it's now `organizationID` from `auth.OrganizationID()`. Those read different headers with different precedence in `dauth` — `UserID()` reads only the deprecated `user-id` headers, while `OrganizationID()` prefers `HeaderNewOrganizationID` and falls back to them.

Any deployment sending both headers now emits a different value under the same key, which breaks dashboards grouping on it. The PR body says "message text and existing fields are unchanged", and the CHANGELOG doesn't mention it. If the rename is intentional, it wants to be a called-out breaking change; if not, keep `auth.UserID()`.

Separately: the same value is keyed `organization_id` on the incoming line (line 100) and `user_id` here. Since the stated goal is joining these two lines downstream, having the same identity under two keys is going to bite whoever writes that query.

firehose/server/blocks.go
```diff
@@ -79,18 +79,77 @@ func (s *Server) Block(ctx context.Context, request *connect.Request[pbfirehose.
...
+			zap.Int("egress_bytes", meter.egressBytes),
+			zap.Duration("duration", time.Since(requestStart)),
+			zap.Error(logErr),
+		}
+		if firstDataSent {
+			fields = append(fields,
+				zap.Duration("time_to_first_data", timeToFirstData),
+				zap.Uint64("resolved_start_block", resolvedStartBlock),
+			)
+		}
```
GabrielCartier: These two fields are omitted when no block was ever sent, but Substreams always emits them zero-valued (`metrics/stats.go`). Since the point is one analytics model over both services, a query doing `avg(time_to_first_data)` gets a different denominator depending on which service it reads — absent rows here, zero rows there. Worth emitting them unconditionally to match.

firehose/server/blocks.go
```diff
@@ -221,6 +278,12 @@ func (s *Server) Blocks(ctx context.Context, request *connect.Request[pbfirehose
 			return NewErrSendBlock(err)
 		}
 
+		if !firstDataSent {
+			firstDataSent = true
+			timeToFirstData = time.Since(requestStart)
+			resolvedStartBlock = block.Number
+		}
```
GabrielCartier, line 288: `resolved_start_block` here is the first block *sent*, which isn't the same thing as the resolved start block. Substreams sets its identically-named field from `requestDetails.ResolvedStartBlockNum`, known before streaming begins.

Two ways they diverge: with `final_blocks_only=true` on a chain whose LIB lags, `stepToProto` skips every non-final step, so the first sent block can be far past the resolved start. And `stepFilter` admits `StepUndo`, so a stream whose first send is an undo reports a number *lower* than the start block.

Given the field name is shared with Substreams, this should come from the resolved start block, not the first send. If you want the first sent block too, that's a separate key.

firehose/server/blocks.go
```diff
@@ -294,65 +357,47 @@ func (s *Server) Blocks(ctx context.Context, request *connect.Request[pbfirehose
 		return err
 	}
 
-	err = str.Run(ctx)
-	meter := getRequestMeter(ctx)
+	runErr = str.Run(ctx)
+	streamRan = true
```
GabrielCartier, line 364: Blocker: this path logs a failure as a success.

If `str.Run` returns nil, `streamRan` is true and `runErr` is nil, so the defer takes `logErr = runErr = nil` and drops the named return entirely. Execution falls through to line 404-405, which logs `"source is not expected to terminate gracefully"` and returns `status.Error(codes.Internal, "unexpected stream completion")`.

So the client gets an `Internal` error, and `"firehose process completed"` carries no `error` field — downstream error-rate analytics counts it clean. This is exactly the case the CHANGELOG's "both lines are now emitted for every request outcome" is meant to cover.

Simplest fix is to set `runErr` on that terminal path before returning, or have the defer prefer the named return when `runErr` is nil.

firehose/server/blocks.go
```diff
@@ -294,65 +357,47 @@ func (s *Server) Blocks(ctx context.Context, request *connect.Request[pbfirehose
...
-		)
-	}
-
-	logger.Info("firehose process completed", fields...)
-	if err != nil {
-		if errors.Is(err, stream.ErrStopBlockReached) {
+	if runErr != nil {
+		if errors.Is(runErr, stream.ErrStopBlockReached) {
 			logger.Info("stream of blocks reached end block")
+			runErr = nil // successful completion, not an error worth surfacing in the completion log
```
GabrielCartier, line 369: This drops the `error` field for stop-block-reached streams. Before this PR the completion line was emitted before the `if err != nil` dispatch, so these logged `error: "stop block reached"`.

Defensible as a deliberate choice — it isn't a failure — but it means log queries can no longer tell "reached its stop block" from "streamed forever and the client left", and both the PR body and CHANGELOG are silent on it. If intentional, the distinction probably wants its own boolean field, and the CHANGELOG should say the `error` value went away.

firehose/server/blocks.go
```diff
@@ -294,65 +357,47 @@ func (s *Server) Blocks(ctx context.Context, request *connect.Request[pbfirehose
...
 
-		var errInvalidArg *stream.ErrInvalidArg
-		if errors.As(err, &errInvalidArg) {
+		if errInvalidArg, ok := errors.AsType[*stream.ErrInvalidArg](runErr); ok {
 			return status.Error(codes.InvalidArgument, errInvalidArg.Error())
 		}
 
-		var errSendBlock *ErrSendBlock
-		if errors.As(err, &errSendBlock) {
+		if errSendBlock, ok := errors.AsType[*ErrSendBlock](runErr); ok {
```
GabrielCartier, line 398: This branch is dead, and has been since before this PR — but the PR rewrote the line and the CHANGELOG now claims client disconnects are properly accounted for, so it's worth fixing here.

`NewErrSendBlock` returns a value `ErrSendBlock`, not a pointer (`firehose/server/errors.go:11`). Nothing in the chain is ever `*ErrSendBlock`, so `errors.AsType[*ErrSendBlock]` can't match. A client disconnecting mid-send returns `ErrSendBlock` from line 278, falls past this check to line 400, and is logged as `"unexpected stream of blocks termination"` and returned as `codes.Internal` instead of the intended `codes.Unavailable`.

Fix is either `errors.AsType[ErrSendBlock]` here or making `NewErrSendBlock` return `*ErrSendBlock`. Also note `ErrSendBlock` has no `Unwrap`, so the inner error isn't reachable via `errors.Is`/`As` either — worth adding while you're in here.

firehose/server/blocks.go
```diff
@@ -294,65 +357,47 @@ func (s *Server) Blocks(ctx context.Context, request *connect.Request[pbfirehose
...
 			// The request is fine, this instance cannot serve it: the client should
 			// retry, here or on another instance, keeping its cursor.
 			return status.Error(codes.Unavailable, errUnavailable.Error())
 		}
 
-		var errInvalidArg *stream.ErrInvalidArg
-		if errors.As(err, &errInvalidArg) {
+		if errInvalidArg, ok := errors.AsType[*stream.ErrInvalidArg](runErr); ok {
 			return status.Error(codes.InvalidArgument, errInvalidArg.Error())
 		}
```
GabrielCartier, line 396: These three `errors.As` → `errors.AsType` conversions are unrelated to log instrumentation and aren't mentioned in the body or CHANGELOG. They look behavior-preserving, but this is the error-classification path — the riskiest part of the file — and it's now churn a reviewer has to verify inside a PR about logging. Would rather see it split out.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you s
 
 ### Added
 
+- The `Blocks` request handler now logs an `"incoming firehose Blocks request"` line as soon as a request starts, carrying `trace_id`, `organization_id`, `api_key_id`, `real_ip`, `start_block`, `stop_block`, `final_blocks_only` and `cursor`. The existing `"firehose process completed"` line gains `duration` (total request time), `time_to_first_data` and `resolved_start_block` (both omitted if no block was ever sent). Both lines are now emitted for every request outcome, including early rejections (session denied, rate limited, unimplemented transforms) and client disconnects, so every request can be paired up downstream by `trace_id`; a normal client cancellation is always logged with `error: "context canceled"`, distinguishing it from a real failure.
```
GabrielCartier: Three claims in this entry don't currently hold:

- "Both lines are now emitted for every request outcome" — the nil-`runErr` fall-through logs no error while returning `Internal` to the client (see comment on line 360).
- "a normal client cancellation is always logged with `error: \"context canceled\"`, distinguishing it from a real failure" — server-initiated cancellation is logged the same way, so it doesn't distinguish (see comment on line 117).
- "existing fields unchanged" is implied but `user_id` changes meaning (see comment on line 142).

Also unmentioned: stop-block-reached streams lose their `error` value.

Worth a re-read of this entry once the code changes land — it's operator-facing and these are the sentences someone will build a dashboard against.

GabrielCartier (changes requested):
Instrumentation is worth having and the shape is right — two lines, deferred completion, pairing by `trace_id`. But three of the body's own claims don't hold once you follow the control flow, and one of them inverts the field this PR exists to add. Details inline; the summary:

- A stream that ends with `Run` returning nil is returned to the client as `Internal` but logged with no `error` field at all.
- The `"context canceled"` normalization reports server-initiated cancellation (quota revocation) as a client disconnect — the one case worth telling apart.
- `user_id` silently changes meaning from `auth.UserID()` to `auth.OrganizationID()`, which the body says is unchanged.

Two things not tied to a diff line:

**`duration` on rate-limited requests includes the deliberate jitter sleep.** The rate-limit path sleeps `rand.Intn(3000)+1000` ms before returning, and that sits inside the deferred timer's window. Rejected requests will show a synthetic ~2.5s latency floor that has nothing to do with server work. Either start the timer after the sleep for that path or subtract it.

**No tests.** `firehose/server/` has no test files at all, and this PR adds a named return read from a defer plus a three-way `(runErr, ctxCause, streamRan)` selection. A table test over those three inputs catches the first two blockers above directly. Worth adding here rather than later, since the whole point is that these lines feed an analytics model — a wrong field is worse than a missing one.

Nit: commit subject has no conventional prefix and runs 71 chars; repo CLAUDE.md asks for `feat:`/`fix:` and the history is mixed but mostly compliant.

firehose/server/blocks.go
maoueh: Yes but we do not keep user_id. We actually change all user_id in logging and other variable names in organization_id, user_id is the wrong info here.

firehose/server/blocks.go
maoueh: Let's just rename `resolved_start_block` to `first_sent_block`, later we will add `resolved_start_block` but requires a bigger changes no necessary here.

Will bridge the differences in dbt later.

firehose/server/blocks.go
maoueh: To investigate though, if claim `can no longer tell "reached its stop block" from "streamed forever and the client left"` is true, agree to change as knowing the distinction is important, but I think client left always result in having a `runErr` (but I could be wrong)

firehose/server/blocks.go
maoueh: We should never mention any cosmetic inner code refactoring, changelog is meant for user knowing what new features there is.

CHANGELOG.md
maoueh: Adjust it after having completed the review updates work

firehose/server/blocks.go
maoueh: Investigated: str.Run() returns a single error, so ErrStopBlockReached and a client-disconnect error (context.Canceled / ErrSendBlock) are mutually exclusive outcomes of the same call — a disconnect before the stop block is reached never surfaces as ErrStopBlockReached. So the empty error field on this path only ever means a clean stop; a genuine disconnect still produces a non-empty error field. Leaving as-is.

firehose/server/blocks.go
maoueh: Kept the errors.As -> errors.AsType conversions for ErrUnavailable/ErrInvalidArg — author preferred keeping them in this PR rather than splitting out. Fixed the real bug in the ErrSendBlock one (it was matching a pointer type that could never occur) since that's a functional regression risk, not cosmetic churn.

sduchesneau (approved):
LGTM

PR #248, opened 2026-09-09
Show the first block of each month in stats-merged-blocks, also use TiB not TB

sduchesneau:
Adds a `START_BLOCK` column to the `stats-merged-blocks` table, second, between `MONTH` and `FILES`: the lowest first-block of the files counted in that month, and the lowest of the whole range on the `TOTAL` row. The JSON buckets carry the same value as `start_block`.

Also, use TiB and GiB instead of TB/GB


joshkauffman:
Pushed 27fb2a0: the new per-bucket field serialized to JSON key `start_block`, which `statsReport` already uses for the `--start-block` filter value — so `report.total.start_block` (observed low block) and `report.start_block` (filter) meant two different things under one name, and the observed value also overlapped `first_block_seen`. Renamed the bucket field to `first_block` (and the table column to `FIRST_BLOCK`). Tests updated, all green.

joshkauffman (approved):

PR #254, opened 2026-09-17
Decode block payloads while reading the node output

sduchesneau:
The reader crashed when a line was longer than `--reader-node-line-buffer-size`: overseer's fixed buffer overflowed, the pipe closed and the node died of a broken pipe. Reading a block also held its line about 3 times in memory: buffer, string copy, decoded payload.

**Changes**
- A new `node-manager/consoleline.Splitter` reads the node output as it arrives and decodes the base64 payload of `FIRE BLOCK` lines on the way. Memory for a block is about 1.5 times its line size (512 MiB block: 1.1 GiB peak instead of 2.2 GiB, and faster).
- The superviser writes the node's stdout/stderr straight into a splitter through the new overseer writer options. `reader-node-stdin` copies stdin into one.
- `firecore.ConsoleReader` receives decoded blocks through the new `mindreader.BlockLineConsoleReader` interface. Chains with their own console reader (fireeth, firenear) keep receiving text lines.
- `--reader-node-line-buffer-size` is now the normal size of the line buffer (default 100 MiB, at least 32 KiB). The buffer grows past it in pieces of that size for a longer line, up to `2861913428` bytes: the longest line whose block, with 1 MiB for the message around it, fits in a 2 GiB response. Each time 50 blocks in a row used less than half of it, it shrinks by half, keeping whole pieces and never going below the normal size.
- Bumps dgrpc so responses up to 2 GiB go through. Request limits are unchanged.

Memory allocations look like this:  each byte goes through base64decode, then each byte out of this is written exactly once in a "piece" (pieces are added as needed), then copied to the final array where it will be processed (sent through stream server and written to disk)

**Performance**

The same output was fed to both reading paths in 32 KiB writes (what `os/exec` does), measured in-process on an Apple M4 Max, 3 runs each:
- **develop:** overseer `OutputStream` with its default 200 MiB buffer, then `base64.StdEncoding.DecodeString` on the line.
- **this PR:** `consoleline.Splitter` with the default 100 MiB buffer.

| Workload | Path | Time | Allocated | Peak heap |
|---|---|---|---|---|
| 200 blocks of 10 MiB (2.6 GiB of output) | develop | 780–854 ms | 7536 MiB | 456–470 MiB |
| | this PR | 656–680 ms | 2100 MiB | 200–220 MiB |
| 20 blocks of 100 MiB (2.6 GiB of output) | develop | 822–885 ms | 7533 MiB | 700 MiB |
| | this PR | 671–687 ms | 2200 MiB | 600 MiB |

- **Time:** about 20% faster.
- **Allocated:** 3.4 to 3.6 times less. develop copies each line about 3 times before decoding (buffer, string, string concatenation), while this PR allocates the decoded payload once, plus the buffer.
- **Peak heap:** it depends on when the garbage collector runs, so allocated bytes are the more reliable number. develop's peak also includes its 200 MiB buffer, allocated up front; a real reader allocates one each for stdout and stderr.
- **Bigger blocks:** develop can't read a block over about 150 MiB with its default buffer. With the buffer raised to 1 GiB, a 512 MiB block peaked at 2.2 GiB against 1.1 GiB with this PR. Running the `firecore` binary on a 256 MiB block, max RSS over 3 runs was 1277 to 1959 MiB on develop (buffer raised to 1 GiB) against 867 to 868 MiB with this PR.

**Depends on**
- streamingfast/overseer#1 (writer options)
- streamingfast/dgrpc#37 (2 GiB responses)

`go.mod` points at those branches' commits and must be updated once they are merged.

**Tested**
- Splitter unit tests compare against line-at-a-time splitting, over many chunk and buffer sizes.
- An end-to-end test sends a 40 MiB block from a real child process through the superviser into `ConsoleReader`.
- `firecore start reader-node-stdin`, run against the same input as develop, including a 60 MiB block, writes identical one-block files.


sduchesneau:
@maoueh do you trust me on that refactor ? ;)

sduchesneau:
@copilot review

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ If you were at `firehose-core` version `1.0.0` and are bumping to `1.1.0`, you s
 
 - Reader: the base64 payload of a `FIRE BLOCK` line is decoded while the line is read, instead of after the whole line was copied into a string. A block no longer crashes the reader when it is bigger than the old 200 MiB line buffer, and reading a block needs about 1.5 times its line size in memory instead of about 3 times (measured on a 512 MiB block: 1.1 GiB peak instead of 2.2 GiB). This applies to chains using the default `firecore.NewConsoleReader`; chains with their own console reader keep receiving `FIRE BLOCK` lines as text.
 - gRPC clients receive responses up to 2 GiB instead of 1 GiB, and servers send responses up to 2 GiB (bumped `dgrpc`), so a block up to 2 GiB goes from the reader through the relayer to Firehose and Substreams. Request limits are unchanged. External clients need their own receive limit raised to get blocks that big. Substreams modules cannot take a block that big: wasm32 has 4 GiB for the block, its decoded form and the output.
+||||||| 78161b0
```
maoueh: What is that?

maoueh (approved):
Discussed testing plan, will run in test-mode on some big networks for a while. If all good, will then test on our networks that have 3 nodes of redudancy for a few days.

If good, will roll out overall the platform.

PR #255, opened 2026-09-17
rpc: add RollingStrategySequential with spread-start option

maoueh:
## Summary
- Adds `RollingStrategySequential`: walks the pool in order from a configurable starting client, rolling forward through the rest for failover. `WithSpreadStart()` spreads that starting client round-robin across successive `WithClients`/`WithClientsContext` calls, so concurrent batched fetches (e.g. `blockpoller`'s parallel block polling) don't all hit the same provider first. Failover order after the start point is unchanged.
- `RollingStrategyAlwaysUseFirst` / `NewRollingStrategyAlwaysUseFirst` are now deprecated: the type is a generic alias for `RollingStrategySequential`, and the constructor is an unchanged, no-option wrapper — existing callers keep compiling with identical behavior.
- Opt-in and non-invasive otherwise: no changes to `Clients`, `WithClientsContext`, `StickyRollingStrategy`, or `blockpoller`. A pool picks `WithSpreadStart()` at construction time — useful for batched/backfill workloads; the default (or sticky) stays right for live tailing, where staying on a known-good provider matters more than spreading load.
- Sequential, failback, and concurrent-selection behavior covered by tests in `rpc/rolling_strategy_test.go` and `rpc/clients_concurrent_test.go`.


sduchesneau (approved):
LGTM

